PR #27, opened 2026-04-12
Handle StreamReplaced event to recover from session conflicts

davidsimoes:
## What

Adds an event handler for `events.StreamReplaced` that signals reconnection after a 30-second delay, reusing the existing reconnect goroutine with exponential backoff.

## Why

When another WhatsApp Web session takes the bridge's slot, whatsmeow emits `StreamReplaced` — not `Disconnected`. The library treats stream-replaced as an "expected" disconnect and suppresses the `Disconnected` event entirely. Since the bridge only handles `Disconnected`, `ConnectFailure`, and `StreamError`, the reconnect goroutine never receives a signal and the bridge sits idle indefinitely until manually restarted.

This is easy to trigger accidentally: opening WhatsApp Web in a browser, the macOS WhatsApp desktop app grabbing the slot, or running two bridge instances with the same session store.

## Design choices

- **30s delay before reconnect** — avoids ping-ponging with the competing client. If the other session was transient (e.g., user briefly opened WA Web), 30s is enough for it to close. The existing backoff (5s → 5min) handles persistent conflicts.
- **Fires in a goroutine** — the event handler must return quickly per whatsmeow's contract; the delay runs in the background.

## Testing

Reproduced by opening a competing WhatsApp Web session while the bridge was connected. Without the patch, the bridge logged `OnDisconnect() called, but it was expected` and went idle permanently. With the patch, it logs the replacement warning, waits 30s, and reconnects successfully.


jack-arturo:
Merged. Thanks for the careful diagnosis on the suppressed Disconnected event — that's the kind of whatsmeow gotcha that costs people days. 🙏

PR #28, opened 2026-04-13
feat: security fixes + feature backports (localhost bind, Unicode search, quoted replies, RFC 3339)

jrkphani:
## Summary

This PR ports a set of security hardening changes and feature improvements developed against the lharries fork onto the VGP codebase.

### Security

- **SEC-1 (Go bridge):** Bind REST API to `127.0.0.1` instead of `0.0.0.0`. The bridge was previously reachable from any host on the LAN — this scopes it to localhost only.
- **SEC-2 (Python server):** Replace `LOWER() + LIKE` with `instr(LOWER(), LOWER())` for all content/name/JID searches. SQLite's `LOWER()` only handles ASCII — Unicode contact names (CJK, Arabic, Devanagari, Telugu, etc.) were silently excluded from search results.
- **SEC-3 (Python server):** Enforce RFC 3339 timestamps with explicit timezone on `after`/`before` filters in `list_messages`. Naive datetimes are now rejected with a clear error; all output timestamps are normalised to UTC with a `Z` suffix.

### Features

- **Quoted reply support:** `send_message` accepts an optional `quoted_id` parameter (message ID from `list_messages` output). The Go bridge looks up the original message in the DB and constructs a proper `ContextInfo` for WhatsApp swipe-reply rendering. Also exposed as `quoted_id` in the MCP tool.
- **Message IDs in output:** `list_messages` now appends `[id:xxx]` to each message line, making it easy to copy an ID for quoting.
- **File captions:** `send_file` accepts an optional `caption` parameter, passed as the media message caption.
- **Explicit document MIME types:** The Go bridge now sets correct MIME types for PDF, DOC/DOCX, XLS/XLSX, PPT/PPTX, CSV, TXT, ZIP (was `application/octet-stream` for all).
- **`FileName` field on documents:** `DocumentMessage` now populates `FileName` in addition to `Title` — some WhatsApp clients display this separately.
- **Safe filepath parsing:** Uses `filepath.Ext` / `filepath.Base` instead of manual `strings.LastIndex` slicing for extension and filename extraction.
- **Interactive/business message text extraction:** `extractTextContent` now handles `InteractiveMessage`, `ButtonsMessage`, `ButtonsResponseMessage`, `ListMessage`, `ListResponseMessage`, and image/video captions — previously these returned empty strings.
- **whatsmeow pinned to `v0.0.0-20260410162419-b95d92207080`** (April 10 2026, newer than the Jan 2026 baseline).

## Test plan

- [ ] Start the Go bridge — confirm it binds to `127.0.0.1:8080` in logs (`Starting REST API server on 127.0.0.1:8080`)
- [ ] Confirm `curl http://0.0.0.0:8080/api/send` fails (connection refused) while `curl http://127.0.0.1:8080/api/send` responds
- [ ] Search contacts using a Unicode name (e.g. Chinese, Arabic, Telugu) — confirm results return
- [ ] Call `list_messages` with `after="2026-01-01"` (no timezone) — confirm error message mentions RFC 3339
- [ ] Call `list_messages` with `after="2026-01-01T00:00:00Z"` — confirm it works and timestamps in output end with `Z`
- [ ] Call `send_message` with a valid `quoted_id` — confirm WhatsApp shows swipe-reply rendering
- [ ] Call `send_file` with a `caption` — confirm caption appears on the media in WhatsApp
- [ ] Send a `.pdf` file — confirm WhatsApp shows it with PDF icon rather than generic document
- [ ] Receive a business/bot interactive message — confirm `list_messages` shows its body text


jack-arturo:
Thanks for putting this together — there's good work here, but the security half of this PR has substantially overlapped with what already shipped in this fork via the `fix/security-localhost-bind-unicode-search` branch (localhost-only bind, Unicode search, RFC 3339 normalization). Merging this as-is would either redo work or risk reverting parts of that pass.

Could you split out a **feature-only** PR with the bits that aren't already in `main`? Specifically the pieces that look novel from a quick read of the diff:

- Quoted/replied message support
- File captions on media sends
- Explicit document MIME types + `FileName` propagation on documents
- Interactive / business message text extraction (so search hits the visible text)

Smaller, scoped PRs are much easier to review and land here, and we can run them through CI individually. If you'd prefer, I can cherry-pick one of the above into a focused PR to start the pattern — let me know which is highest value to you.

(Marking with `needs:split` and leaving open for now.)

jack-arturo:
Closing this stale umbrella PR. The release branch has already landed the security and feature pieces from this in smaller PRs, including localhost/Unicode/RFC3339 hardening, document MIME/FileName handling, caller logging, and quoted replies.

Any remaining behavior from this branch should come back as a new focused PR against current `main`, with one concern per PR.

PR #40, opened 2026-04-18
fix(bridge): include message ID in media filenames to prevent same-second collisions

edmenendez:
## Summary

Two media messages arriving in the same second are silently corrupted today because the bridge names all media files using only `<type>_YYYYMMDD_HHMMSS.<ext>`. When the second message arrives, the bridge's `os.Stat` existence check at `downloadMedia` finds the first message's file and skips the download — so the second message's content never lands on disk. Any later re-download attempt (via `/api/download` or a backfill tool) then transcribes/serves one message's content under another message's ID.

This is particularly easy to hit with voice-note bursts (two quick `🎙️` messages in a group, or a bulk forward), but applies to all media types — image, video, audio, document.

## Reproduction

1. Send two voice notes to the same chat within the same second (easy in group chats or via a send-twice script).
2. Observe `store/<chat>/audio_YYYYMMDD_HHMMSS.ogg` is written only once (the first arrival).
3. Request `/api/download` for the second message's ID → returns the first message's file content, silently labeled as the second message.

## Fix

Embed the WhatsApp message ID in the filename:

- Before: `audio_20260418_094934.ogg`
- After:  `audio_20260418_094934_3A88F4A40794D5396EE4.ogg`

Changes:

1. `extractMediaInfo` now takes `msgID` and suffixes it after the timestamp for image/video/audio/document.
2. Both callers (`handleMessage` and `handleHistorySync`) pass the message ID they already have in hand.
3. `downloadMedia` reconstructs the new filename so the `os.Stat` existence check matches what `extractMediaInfo` emits.

Scope: ~10 lines of non-trivial change plus threading a parameter. No behavior change for messages arriving in distinct seconds; the only observable difference is longer filenames.

## Migration for existing installs

Older installs will still have legacy `<type>_YYYYMMDD_HHMMSS.<ext>` files on disk. The new bridge will happily keep using those via `os.Stat` skipping the download if it happens to match the new expected name — but in practice, new-format and legacy-format files won't match, so the bridge will re-download once per message. That's wasteful but safe.

For a clean-state migration (rename legacy files to the new format and update `messages.filename`), we have a Python script we're willing to contribute in a follow-up PR if useful.

## Test plan

- [x] Two voice notes sent in the same second each land under their own filename
- [x] `/api/download` for each message returns distinct content
- [x] `go build ./...` passes, `go vet ./...` clean
- [x] Running personal install on the fix branch for ~30 minutes of live message flow, no regressions


jack-arturo:
Merged. 🙏

PR #41, opened 2026-04-18
fix(bridge): auto-download runs after StoreMessage to avoid lookup race

edmenendez:
## Summary

`handleMessage` launches the auto-download goroutine **before** inserting the message row into SQLite. `downloadMedia`'s first action is `SELECT ... FROM messages WHERE id=? AND chat_jid=?` — so when the goroutine wins the race, the query returns no rows and the download is silently abandoned.

In practice this fires often on fast-arriving media (two voice notes in the same conversation within 1–3 seconds, or a burst of status broadcasts). Every time, one or more media messages are dropped and only the DB row remains. The file never lands on disk unless a later tool explicitly calls `/api/download` for that message ID.

## Reproduction

1. Send two voice notes to yourself within ~3 seconds.
2. Observe the bridge log:
   ```
   [Client INFO] Auto-downloading audio media for message A...
   [Client WARN] ❌ Auto-download failed: failed to find message: sql: no rows in result set
   ```
3. The file for the losing message never appears in `store/<chat>/`.

I hit this consistently on my Mac with the bridge running against my own phone. Four of five same-burst media messages hit the race across today's testing.

## Fix

Reorder the two operations in `handleMessage`:

- **Before:** spawn goroutine → (goroutine queries DB) → call `StoreMessage`
- **After:**  call `StoreMessage` → spawn goroutine → (goroutine queries DB)

Also: only spawn the download goroutine when `StoreMessage` returned no error — no point fetching media for a row that failed to persist.

## Test plan

- [x] `go build ./...` passes, `go vet ./...` clean
- [x] Running patched bridge locally for 10 minutes of live traffic — zero "failed to find message" warnings where previously there were several
- [x] Send two voice notes in ~1 second → both land on disk under their own filenames (in conjunction with #40 for the filename uniqueness)


jack-arturo:
Merged. Clean diagnosis and a tight diff — exactly the kind of contribution that makes maintaining this fork easy. 🙏

Heads up: I'm planning to add you as a CODEOWNER for `whatsapp-bridge/**` and offer triage rights on the repo if you're up for it. No obligation — just means your name shows up on bridge PRs and you can label/close issues. Let me know.

PR #42, opened 2026-04-18
fix(bridge): surface image/video/document captions in extractTextContent

edmenendez:
## Summary

Surface user-typed captions on `ImageMessage`, `VideoMessage`, and `DocumentMessage` when extracting message content. These fields were silently dropped — on a production `messages.db` with 3,869 image rows, every single one has empty `content`, even though many of them carry captions in WhatsApp.

After this change, a photo sent with "look what the kids drew" in the caption field stores that text in `messages.content`, making it searchable alongside plain-text messages.

## Change

`extractTextContent` (`whatsapp-bridge/main.go`) falls through to each media type's `GetCaption()` after the existing `Conversation` / `ExtendedTextMessage` checks:

```
Conversation          → use
ExtendedTextMessage   → use
ImageMessage          → use GetCaption()   ← new
VideoMessage          → use GetCaption()   ← new
DocumentMessage       → use GetCaption()   ← new
(else)                → ""
```

`AudioMessage` doesn't carry a caption field in the WhatsApp protocol, so it's not included.

## Testing

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./...` — all existing tests pass
- [x] New table-driven unit test `TestExtractTextContent_SurfacesMediaCaptions` covers:
  - Plain `Conversation` (regression)
  - `ExtendedTextMessage` (regression)
  - `ImageMessage` with caption
  - `VideoMessage` with caption
  - `DocumentMessage` with caption
  - `ImageMessage` without caption → returns `""`
  - Nil message → returns `""`

## Scope

Intentionally narrow — only the extraction function is changed. Doesn't touch storage, download, or any of the MCP-side code. Once this is merged, existing MCP search tools automatically pick up the newly-populated captions via the existing `messages.content` column; no schema change needed.


jack-arturo:
Merged. 🙏

PR #43, opened 2026-04-18
fix(mcp): match messages by both phone number and LID via whatsmeow_lid_map

edmenendez:
## Summary

`list_messages`, `get_contact_chats`, and `get_last_interaction` all match on `messages.sender`, but that column holds either a bare phone number (e.g. `13232432100`) or a bare LID (e.g. `231241139937355`) depending on the wire message. A filter keyed on phone number silently misses every message whose sender was stored under the LID form — entire voice-note threads disappear from query results even though the rows are there on disk.

Resolve the counterpart via `whatsmeow_lid_map` (already populated by the bridge in `whatsapp.db`) and match `sender IN (pn, lid)` in all three query sites.

## Reproduction

1. Send two voice notes from your paired phone to your self-chat.
2. Confirm the rows land in `messages.db` with `sender = '231241139937355'` (your LID), `chat_jid = '13232432100@s.whatsapp.net'`.
3. Call `list_messages(sender_phone_number="13232432100", limit=5)` — the new rows are absent.
4. Inspect `whatsmeow_lid_map` in `whatsapp.db` and confirm the LID↔PN pair already exists.

## Change

New helper `_sender_aliases(value)`:

- Strips any `@…` suffix → `bare`
- Looks up the counterpart in `whatsmeow_lid_map` (PN → LID, falling back to LID → PN)
- Returns `[bare]` (or `[bare, counterpart]` when a mapping exists)
- Falls through cleanly if `whatsapp.db` or the table doesn't exist

Three call sites updated to use `IN (...)` over the alias list instead of `sender = ?`:

- `list_messages` — filter on `sender_phone_number`
- `get_contact_chats` — `m.sender = ?` branch of the WHERE
- `get_last_interaction` — same

Also adds the `WHATSMEOW_DB_PATH` constant (same env var and default as used on `fix/contact-sync-from-whatsmeow`).

## Test plan

- [x] `pytest` — 12/12 pass
- [x] `python -c "import ast; ast.parse(open('whatsapp-mcp-server/whatsapp.py').read())"` clean
- [ ] Restart the MCP server and re-run `list_messages(sender_phone_number="…")` against a contact known to have LID-stored messages; confirm previously-missing rows now return
- [ ] Regression: a contact with no LID mapping returns the same results as before (`aliases = [pn]` → unchanged SQL behavior)
- [ ] Regression: `get_contact_chats` and `get_last_interaction` on the same contact now include group-chat messages whose sender is stored as the LID


jack-arturo:
Merged. 🙏

PR #75, opened 2026-04-28
chore(security): defense-in-depth hardening

Poker71:
# chore(security): defense-in-depth hardening

## What

A small, additive defense-in-depth pass on the bridge and MCP server. No
behavior change for the default-loopback setup. Each item is a discrete fix
that could ship independently — bundled here per CONTRIBUTING.md "one concern,
≤300 LOC" because they all live in the same security-hygiene category.

## Why

Found while reading through the codebase end-to-end. Nothing here is an active
exploit, but several are footguns that would bite someone who deviates from the
default loopback config (LAN setups, multi-user dev hosts, log aggregators
ingesting stdout). All fit in the "Security" bucket of `ROADMAP.md`:

> Bind defaults that fail safe (localhost first). Auditing of file/path inputs.

### 1. `whatsapp-bridge`: stop logging message bodies on `/api/send`

`fmt.Println("Received request to send message", req.Message, ...)` wrote the
full message body to stdout on every send. Anyone tailing the bridge log (or
piping it into a log aggregator / launchd output file) would end up with a
plaintext copy of every outgoing WhatsApp message — including pasted
secrets/OTPs/tokens. Replaced with metadata-only logging
(`recipient`, `message_len`, `has_media`).

### 2. `whatsapp-bridge`: path-traversal guard on media storage

`downloadMedia` built `chatDir := "store/" + chatJID` with only `:` replaced.
`chatJID` is normally well-formed (whatsmeow-produced), but it ultimately
originates from the remote WhatsApp protocol. Defensively `filepath.Clean`-ing
+ asserting the resolved path stays inside `store/` removes a whole class of
"crafted JID escapes the storage dir" risk for free.

### 3. `whatsapp-bridge`: validate `WEBHOOK_URL` at startup

Previously a misconfigured `WEBHOOK_URL` (typo, missing scheme, or — more
importantly — a non-loopback `http://` URL) would silently succeed and stream
chat content + base64 image data plaintext over the network. Now:

- Scheme must be `http` or `https`.
- Host must be present.
- Non-loopback hosts require `https://`. Override with
  `WEBHOOK_ALLOW_INSECURE=true` (loud warning logged) for trusted-LAN
  receivers like Home Assistant.
- Validation runs once at first delivery; failures disable webhook output and
  log a single clear error rather than spamming on every message.

Includes `webhook_test.go` covering the matrix.

### 4. `whatsapp-bridge`: defensive response headers

Adds `X-Content-Type-Options: nosniff`, `Cache-Control: no-store`, and
`Referrer-Policy: no-referrer` to every REST response via a tiny
`withSecureHeaders` wrapper. Doesn't change functionality; reduces the
blast-radius of any local browser tab that bypasses same-origin and pokes at
`127.0.0.1:8080`.

### 5. `whatsapp-mcp-server`: pin ffmpeg resolution

`audio.py` invoked `ffmpeg` via PATH lookup, so a user-writable directory
earlier on `PATH` (e.g. a poisoned `~/.local/bin/ffmpeg`) silently took over
audio conversion. Added `_resolve_ffmpeg()`: prefers `FFMPEG_PATH` (must be an
absolute, executable file) and falls back to `shutil.which("ffmpeg")` with a
clear error if neither resolves.

## Out of scope (intentionally)

Mentioned only so reviewers know I considered them and chose not to bundle:

- **REST-API authentication** — proper bearer-token / Unix-socket fix needs
  design discussion (touches "authentication or network bind defaults" per
  CONTRIBUTING.md). Will open a separate issue / private security advisory.
- **`messages.db` at-rest encryption (SQLCipher)** — schema-impacting,
  warrants its own design discussion.
- **Rate-limiting `/api/send`** — needs a config story for the right limits
  and per-recipient vs global scope.

## Test plan

- [x] `uv run pytest -q` → 12 passed
- [x] `uv run ruff check .` → clean
- [x] `uv run ruff format --check .` → clean
- [ ] `golangci-lint run` (CI)
- [ ] `go build ./...` (CI)
- [ ] `go test ./...` (new `TestValidateWebhookURL` cases) (CI)

## Risk / rollback

Every change is purely additive on the default-loopback config — same defaults,
same behavior. The webhook validator can refuse to deliver if `WEBHOOK_URL` is
malformed, which is a *behavioral* change for misconfigured deploys — but
silent delivery to a typo'd URL was a footgun, not a feature. Rollback is
revert-the-PR; no migration state.

### DIFF
--- a/whatsapp-bridge/main.go
+++ b/whatsapp-bridge/main.go
@@ -1296,8 +1296,28 @@ func downloadMedia(client *whatsmeow.Client, messageStore *MessageStore, message
 	}
 	filename := fmt.Sprintf("%s_%s_%s%s", mediaType, timestamp.Format("20060102_150405"), messageID, ext)
 
-	// First, check if we already have this file
-	chatDir := fmt.Sprintf("store/%s", strings.ReplaceAll(chatJID, ":", "_"))
+	// Build the per-chat storage directory under store/.
+	//
+	// chatJID is normally produced by whatsmeow and well-formed, but it
+	// originates upstream from the WhatsApp protocol and ends up in a
+	// filesystem path here. We sanitize it defensively so a malformed or
+	// crafted JID can never escape the store/ directory via "../" or
+	// absolute-path components.
+	safeChatJID := strings.ReplaceAll(chatJID, ":", "_")
+	chatDir := filepath.Join("store", safeChatJID)
+	storeAbs, err := filepath.Abs("store")
+	if err != nil {
+		return false, "", "", "", fmt.Errorf("failed to resolve store path: %v", err)
+	}
+	chatDirAbs, err := filepath.Abs(chatDir)
+	if err != nil {
+		return false, "", "", "", fmt.Errorf("failed to resolve chat dir path: %v", err)
+	}
+	// Ensure the resolved chatDir is strictly inside store/ (defence in depth
+	// against future refactors or unexpected JID formats).
+	if chatDirAbs != storeAbs && !strings.HasPrefix(chatDirAbs, storeAbs+string(os.PathSeparator)) {
+		return false, "", "", "", fmt.Errorf("invalid chat JID: resolved path escapes store directory")
+	}
 
 	// Create directory for the chat if it doesn't exist
 	if err := os.MkdirAll(chatDir, 0755); err != nil {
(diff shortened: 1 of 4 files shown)


Poker71:
Closing temporarily to coordinate privately with the maintainer on a related security topic. Will reopen once the coordination is resolved. No issue with the changes themselves — purely a sequencing decision. Apologies for the noise.

PR #76, opened 2026-04-30
Persist outbound messages to local SQLite after SendMessage

nicoacn:
## Bug

`sendWhatsAppMessage` calls `client.SendMessage()` and returns success, but never persists the message in the local SQLite. Outbound messages are invisible to `list_messages` / `get_last_interaction` (and to the `last_message` field of `list_chats`) until WhatsApp eventually propagates them via multi-device sync from another linked device — which can take seconds, minutes, or never (if no other device is connected).

This breaks any use case that depends on tracking what the bridge itself just sent: logging, audit trails, or chat-driven agents that need to see their own outbound replies in subsequent reads.

## Reproduction

1. Run the bridge.
2. `POST /api/send` with a recipient + message.
3. Response: `{"success": true, "message": "Message sent to ..."}` — the message arrives at the recipient.
4. Query `list_messages` filtered by the chat JID. The message you just sent is **not** there.

## Root cause

Two parts:

1. After `client.SendMessage()` returns OK, `sendWhatsAppMessage` returns without calling `messageStore.StoreMessage()`.
2. whatsmeow does not re-emit `*events.Message` for messages this client itself sent via `SendMessage`. The event handler in `main.go` only stores incoming events.

Combined, outbound is never persisted locally.

## Fix

After `client.SendMessage` succeeds, explicitly call `messageStore.StoreChat` + `messageStore.StoreMessage` with `IsFromMe=true` and the response's `ID` / `Timestamp`.

Two subtleties:

- **Pre-LID-resolution JID**: the function resolves `recipientJID` to a `@lid` form before sending (for migrated contacts). Storing under that LID would fragment the chat — incoming events come back via `handleMessage` which normalizes them to `@s.whatsapp.net` via `resolveLIDChat`. Captured `storageJID := recipientJID` *before* the LID resolution block to keep both directions in the same chat.
- **Function signature**: `sendWhatsAppMessage` now takes `*MessageStore`. Only one caller (`/api/send` handler) — updated.

## Limitations

For media attachments, only `mediaType` and `filename` are persisted. `URL`, `MediaKey`, `FileSHA256`, etc. would require capturing the `client.Upload` response earlier in the function and threading it down. Left as a follow-up since `download_media` over outbound has limited value (the caller already has the local file at `mediaPath`).

## Test plan

- [x] Build green: `go build` in `whatsapp-bridge/`.
- [x] Send a text message via `/api/send`. Verify it appears in `list_messages` filtered by the phone JID immediately, with `is_from_me=1`.
- [x] Verify `get_last_interaction` returns the new message.
- [x] Verify the chat JID stays under `@s.whatsapp.net` (not split between `@s.whatsapp.net` and `@lid`) for LID-migrated contacts.

In production use since 2026-04-30.

### DIFF
--- a/whatsapp-bridge/main.go
+++ b/whatsapp-bridge/main.go
@@ -667,7 +667,7 @@ type SendMessageRequest struct {
 }
 
 // Function to send a WhatsApp message
-func sendWhatsAppMessage(client *whatsmeow.Client, recipient string, message string, mediaPath string) (bool, string) {
+func sendWhatsAppMessage(client *whatsmeow.Client, messageStore *MessageStore, recipient string, message string, mediaPath string) (bool, string) {
 	if !client.IsConnected() {
 		return false, "Not connected to WhatsApp"
 	}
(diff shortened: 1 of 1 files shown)


jack-arturo:
@copilot apply changes based on the comments in [this thread](https://github.com/verygoodplugins/whatsapp-mcp/pull/76#pullrequestreview-4255707522)

PR #84, opened 2026-05-07
fix(bridge): store outbound messages in local DB at send time

amazingpanama:
## Summary

`/api/send` previously delivered messages via whatsmeow but never called `StoreMessage`, leaving outbound rows missing from the SQLite store until the next HistorySync on reconnect backfilled them. In practice many outbound messages never backfilled at all, so they were missing from local queries.

This change has `sendWhatsAppMessage` store the sent message immediately after a successful `client.SendMessage`, using the pre-LID-resolution JID so rows align with how inbound messages and HistorySync normalize `chat_jid`.

## Changes

- `sendWhatsAppMessage` now takes a `*MessageStore` parameter and stores the outbound message after a successful send
- Captures the chat JID before LID resolution, so storage matches inbound + HistorySync conventions
- Uses `sendResp.ID` and `sendResp.Timestamp` from whatsmeow's response so the stored row matches what WhatsApp acknowledged
- Logs failures as warnings (storage failure does not fail the send)
- gitignore: adds `whatsapp-bridge.bak-*` and `whatsapp-bridge.new` (safety copies created during hot-swap of the binary)

## Why this matters

Without this fix, applications consuming the local SQLite store cannot rely on it as the source of truth for outbound messages — they have to wait for HistorySync, which is unreliable. With this fix, outbound messages are immediately visible to any consumer reading the DB.

## Test plan

- [x] Send a personal message via /api/send → row appears in messages table immediately with from_me=1
- [x] Send a group message → row appears with the group JID (not the LID-resolved form)
- [x] Send fails (e.g. recipient not migrated) → no orphaned row in DB
- [x] Storage failure does not break the send response

## Notes

Running this in production for ~2 weeks against a WhatsApp Business account with no issues. Happy to adjust style or split commits if helpful.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,8 @@ package-lock.json
 # Go binaries
 whatsapp-bridge/whatsapp-bridge
 whatsapp-bridge/whatsapp-client
+whatsapp-bridge/whatsapp-bridge.bak-*
+whatsapp-bridge/whatsapp-bridge.new
 *.exe
 
 # WhatsApp data (personal/sensitive)
--- a/whatsapp-bridge/main.go
+++ b/whatsapp-bridge/main.go
@@ -667,7 +667,7 @@ type SendMessageRequest struct {
 }
 
 // Function to send a WhatsApp message
-func sendWhatsAppMessage(client *whatsmeow.Client, recipient string, message string, mediaPath string) (bool, string) {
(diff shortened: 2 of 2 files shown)


amazingpanama:
Closing this PR. Will keep the fix local for now.

PR #92, opened 2026-05-14
feat(bridge): caller-fingerprinting on /api/send for downstream-producer identification

dpgvan:
## What

Adds one log line at the top of the \`/api/send\` handler capturing \`RemoteAddr\` + \`User-Agent\` on every incoming request. Identifies which downstream process is producing relay-error messages.

## Why

In our deployment, an unidentified downstream process has been calling \`/api/send\` for 5+ weeks with generic "Something went wrong" error bodies paired 1:1 with PIL alerts (~60-90s lag). We've been unable to identify the calling process because the bridge logs the request body but not the client identity.

Caller-fingerprinting is the minimum addition that resolves this — surfaces \`RemoteAddr\` + \`User-Agent\` so we can cross-reference with \`lsof -i :8080\` and \`ps aux\` to identify the source. Privacy-conscious: only network identifiers + UA are logged; request body content stays as-is in the existing \`Received request to send message\` log line (unchanged).

## Testing

Synthetic send:
\`\`\`
\$ curl -s -X POST http://localhost:8080/api/send \\
    -H "User-Agent: test-relay-producer-id" -d '{}'
\`\`\`

Log line in stdout:
\`\`\`
[/api/send] from=[::1]:57305 ua="test-relay-producer-id"
\`\`\`

Works — fingerprint logged before body validation (malformed requests still get identified).

## Notes

- 1-line addition, zero behavior change for legitimate clients
- Doesn't touch other \`/api/*\` handlers (out of scope; same pattern could apply if useful)
- Builds clean against current main; go.mod/go.sum unchanged

Happy to adjust the log format if it doesn't match your conventions.

### DIFF
--- a/whatsapp-bridge/main.go
+++ b/whatsapp-bridge/main.go
@@ -403,6 +403,73 @@ func (store *MessageStore) GetMessages(chatJID string, limit int) ([]Message, er
 	return messages, nil
 }
 
+// DeleteMessage removes a single message row from the local store by its ID.
+// GTM-1204: "delete-for-me" cleanup for WhatsApp-to-self spam. Does NOT send a
+// revocation to WhatsApp servers — the message stays in other participants'
+// views. Suited for self-chat noise cleanup where "delete-for-everyone" is a
+// no-op (only one participant).
+//
+// Returns the number of rows deleted (0 or 1) and any SQL error.
+func (store *MessageStore) DeleteMessage(id, chatJID string) (int64, error) {
+	res, err := store.db.Exec(
+		"DELETE FROM messages WHERE id = ? AND chat_jid = ?",
+		id, chatJID,
+	)
+	if err != nil {
+		return 0, err
+	}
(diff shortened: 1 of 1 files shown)


jack-arturo (commented):
Agreed on the LIKE wildcard issue. I’m not planning to merge this PR as-is for the patch update because it mixes the caller fingerprint log with delete endpoints, mention/MIME changes, and GFV-specific cleanup context.

I extracted the narrow `/api/send` caller fingerprint logging into #96, which logs only remote address and user agent and does not log message bodies. If we revisit the delete endpoint work separately, the bulk delete prefix filter should escape `%` / `_` and use an `ESCAPE` clause or a non-LIKE strategy before it’s mergeable.

jack-arturo:
Closing this as superseded by #96, which landed the caller identity logging for `/api/send` as a focused fix and is already on `main`. Thanks for the original report and implementation direction.

PR #103, opened 2026-05-18
fix(bridge): persist outgoing messages on send instead of relying on events.Message echo

### LINKED ISSUE #102: Outgoing messages sometimes vanish from messages.db: send API reports success but list_messages cannot find them later
## Summary

`sendWhatsAppMessage` (in `whatsapp-bridge/main.go`) only delegates to `client.SendMessage` and reports success. Local persistence in `messages.db` depends entirely on the `events.Message` echo callback being fired for the outgoing message. When that callback does not arrive — which happens in practice — the message is delivered to the recipient but never lands in the bridge's local store. `list_messages` then returns a false negative.

## Impact

Downstream MCP clients (Claude / Cursor) cannot reliably reason about conversation state because outgoing messages they themselves just sent may be invisible. In our deployment we hit this at least twice in one week of normal use; the recipient sees the message in WhatsApp but `list_messages` reports nothing.

## Reproduction

1. Set up the bridge per the README, pair a device, wait for `connected: true`.
2. Send a message via the MCP `send_message` tool or directly:
   ```bash
   curl -X POST http://localhost:8080/api/send \
     -H 'Content-Type: application/json' \
     -d '{"recipient":"<phone>","message":"hello"}'
   ```
3. Verify the recipient received it on their device.
4. Query the local store:
   ```bash
   sqlite3 store/messages.db \
     "SELECT id, chat_jid, timestamp, content FROM messages WHERE content='hello' AND is_from_me=1;"
   ```

Some percentage of sends will return no row, despite the API reporting success.

## Root cause

`sendWhatsAppMessage` discards the `SendResponse` from `client.SendMessage` and never writes to the local store. Persistence is delegated exclusively to the `*events.Message` handler in `handleMessage`. WhatsApp's fan-out of outgoing messages to companion devices is best-effort — the echo may be dropped during reconnects, rapid bursts, server-side dedup, or other transient conditions.

## Fix proposed

See accompanying PR. Persist the outgoing message in the send path itself, using `SendResponse.ID` / `SendResponse.Timestamp` plus the existing extractors (`extractTextContent`, `extractMediaInfo`, `GetChatName`, `StoreChat`, `StoreMessage`). `StoreMessage` already uses `INSERT OR REPLACE`, so the change is idempotent with the echo path — if the event arrives later it harmlessly overwrites with identical data; if it never arrives, the row is already there.

## Environment

- macOS 26 / Apple Silicon
- Go 1.26.2, whatsmeow `v0.0.0-20260107124630-ccfa04f8e445`
- whatsapp-mcp `main` @ `d362557`

jmhmunya:
Closes #102

## Problem

`sendWhatsAppMessage` only delegates to `client.SendMessage` and reports success. Local persistence in `messages.db` depends entirely on the `events.Message` echo callback being fired for the outgoing message. When that callback does not arrive — which happens in practice for sends originating from this bridge — the message is delivered to the recipient but never lands in the local store. `list_messages` then returns a false negative.

We hit this at least twice in one week of normal use; the recipient sees the message in WhatsApp but `list_messages` reports nothing, leading downstream MCP consumers (Claude / Cursor) to incorrect conclusions about conversation state.

## Solution

Persist outgoing messages in the send path itself, immediately after `client.SendMessage` returns successfully:

1. **`sendWhatsAppMessage`** now receives `messageStore *MessageStore` and captures the `SendResponse` (`resp.ID`, `resp.Timestamp`) which were previously discarded.
2. **New `persistOutgoingMessage` helper** writes the row to `messages.db` reusing the same extractors as `handleMessage` (`extractTextContent`, `extractMediaInfo`, `GetChatName`, `StoreChat`, `StoreMessage`) — so field semantics stay identical to the echo path.
3. **`originalJID` (pre-LID resolution)** is used for storage, matching `handleMessage`'s normalisation via `resolveLIDChat`. This avoids creating duplicate chat rows for the same contact (one under `<phone>@s.whatsapp.net`, another under `<lid>@lid`).
4. Callsite in `http.HandleFunc("/api/send", …)` updated to pass `messageStore`.

## Race-safety

`StoreMessage` already uses `INSERT OR REPLACE`, so if the echo event eventually arrives and `handleMessage` runs, it harmlessly overwrites the row with identical data. If the echo never arrives, the row from the send path is already there. Idempotent in both directions.

## Testing

Verified with two real outgoing sends on 2026-05-18 to a test contact. Both persisted correctly under `<phone>@s.whatsapp.net` with `is_from_me=1`:

| id | chat_jid | timestamp | is_from_me |
|---|---|---|---|
| `3EB0C0308034FA91359764` | `34666247572@s.whatsapp.net` | `2026-05-18 15:58:46+02:00` | 1 |
| `3EB00D0DB8B1B140C800B7` | `34666247572@s.whatsapp.net` | `2026-05-18 16:00:01+02:00` | 1 |

Both rows landed under the PN `chat_jid` (not LID), confirming the resolution fix works.

## Notes

- No new dependencies.
- No public API change. `sendWhatsAppMessage` is package-private; its signature changes from `(client, recipient, message, mediaPath)` to `(client, messageStore, recipient, message, mediaPath)`. The only caller is the `/api/send` HTTP handler in the same file, updated in this PR.
- Errors during local persistence are logged via `fmt.Printf` and do not affect the send response — consistent with how other warnings are emitted in `sendWhatsAppMessage`.


jack-arturo:
Thanks for the detailed report and PR @jmhmunya 

I rechecked this against current `main`. The issue’s text-message reproduction is already fixed by #76 (`9841e17`), which persists outbound messages immediately after `SendMessage` using the returned message ID/timestamp and stores them under the normalized chat JID. That shipped in `v0.2.1` on 2026-05-14.

The issue was filed against `main @ d362557`, which predates #76, so I think this was a stale-checkout/version mismatch rather than missing work in current `main`.

One thing #103 does surface: outbound media persistence could be made richer. Current `main` records outbound media type/filename but not the uploaded media URL/key/hash metadata. That is adjacent to this issue, but not needed for the `list_messages` false-negative repro and should be handled as a separate focused PR so we preserve the newer `main` behavior around chat-name preservation, media path hardening, disappearing-message settings, and JID resolution.

I’m closing this PR because, after resolving conflicts, it has no remaining diff and CI is green. If you want to open a follow-up for outbound media metadata parity, that would be welcome.

PR #110, opened 2026-05-23
feat(bridge): add support for stickers

tobiaspal:
Tested locally with both inbound and outbound stickers: rows appear in `messages` with `media_type='sticker'`, `.webp` files land under the per-chat store directory, and the v0.2.1 REVOKE handler (#99) stamps `deleted_at` on the row when the sender deletes a stored sticker.


jack-arturo:
Deferring stickers from today’s v0.3.0 release so the release can stay focused on already-green fixes and small MCP bug fixes. This remains a reasonable bridge event-coverage feature, but it needs normal review + CI before landing.

jack-arturo (approved):
Verified against current main: Go tests, build, and golangci-lint pass locally after refreshing the branch.

PR #131, opened 2026-06-10
feat: expose pairing QR + programmatic relink (start REST server before pairing)

squarestech25:
## Summary

Lets a local UI or automation re-link the WhatsApp device **without terminal access**, by exposing the pairing QR over the (already loopback + token-gated) REST API and making the server reachable during pairing.

### New endpoints
- `GET /api/qr` — returns `{logged_in, connected, code, updated}`. `code` is the live whatsmeow pairing string while unlinked (empty once paired).
- `POST /api/logout` — unlinks the current device, then exits so the supervisor (launchd `KeepAlive`) restarts the bridge unlinked and a fresh QR is emitted. `?archive=true` first moves `messages.db` (+ WAL/SHM) into `store/archive/messages-<ts>.db` so a newly linked number starts with a clean history instead of mixing.

### Why the startup change
`/api/qr` is only meaningful **before** login, but the REST server previously started **only after** a successful login — so during the entire QR window the server (and `/api/health`) was unreachable and nothing could render the code. This moves `startRESTServer` to run before the connect loop (it already runs in its own goroutine).

### Continuous QR
whatsmeow's QR channel emits a finite batch (~100s) then times out. The unlinked branch now requests a fresh channel when a batch expires (publishing each code to `/api/qr`) instead of giving up after one batch, so a scannable code is always available until the user pairs.

## Security
Both endpoints use the existing `withAuth` wrapper (bearer token + loopback `Host` allow-list). `/api/logout` is destructive and POST-only. `/api/qr` exposes the short-lived device-linking code only while unlinked. No change to the bind address (still `127.0.0.1`).

## Testing
- `go build`, `go vet`, and `go test ./...` all pass.
- Verified end-to-end on macOS: unlink via `/api/logout` → bridge restarts → `/api/qr` serves a rotating code while unlinked → scanning re-links successfully → `/api/health` reports `connected:true`.

## Notes
- Touches auth/network-exposed surface, so flagging per AGENTS.md. Happy to split the "start REST server before pairing" fix into its own PR if preferred.

### DIFF
--- a/whatsapp-bridge/main.go
+++ b/whatsapp-bridge/main.go
@@ -17,6 +17,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jack-arturo:
Closing this for the v0.4.0 queue rather than shipping it as-is. The underlying idea is useful, but this PR combines several separate high-scrutiny changes: `/api/qr`, destructive `/api/logout`, optional message DB archiving, continuous QR regeneration, and moving REST startup before pairing. Please split this into smaller PRs with tests and security notes for each behavior; the safest starting point is the non-destructive early REST/health or QR-read surface by itself.

PR #134, opened 2026-06-15
fix(bridge): extract text from Template, Button, Interactive, and List messages

nakulgupta-lab:
## Problem

`extractTextContent` in `whatsapp-bridge/main.go` only handled `GetConversation()`, `GetExtendedTextMessage()`, and media captions. Several common WhatsApp Business message types silently returned `""`:

- **TemplateMessage** — WABA/BSP-sent template notifications
- **ButtonsMessage** — quick-reply button messages  
- **InteractiveMessage** — list/button interactive messages
- **ListMessage** — list picker messages
- **ButtonsResponseMessage** / **TemplateButtonReplyMessage** — user replies to above

When `extractTextContent` returned `""` and `mediaType` was also `""`, the storage gate (~line 1521) early-returned without inserting the row into SQLite. All inbound/outbound template messages were silently dropped from the message store.

## Fix

Added handler branches for each missing type in `extractTextContent`:

| Type | Field extracted |
|------|----------------|
| `TemplateMessage` | `HydratedTemplate.HydratedContentText` |
| `ButtonsMessage` | `ContentText` → fallback `Text` |
| `InteractiveMessage` | `Body.Text` |
| `ListMessage` | `Description` |
| `ButtonsResponseMessage` | `SelectedDisplayText` |
| `TemplateButtonReplyMessage` | `SelectedDisplayText` |

## Testing

Verified on a live WhatsApp Business account (WABA Connect BSP):
- Template message → landed in SQLite `messages` table with full body extracted
- Existing conversation/extended-text paths unaffected
- Bridge rebuilt and restarted cleanly with no regressions

## Notes

Each branch only fires if the message type is non-nil and extracted string is non-empty — cannot interfere with existing handled types.


jack-arturo (approved):
Finished the release-scoped changes: removed the unrelated pair-phone flag, added focused extraction tests, and verified Go tests/build/lint locally.

PR #135, opened 2026-06-15
fix(bridge): fall back to local whatsmeow_contacts for chat name resolution

evandrodevbr:
## Problem

When a new individual chat appears, `GetChatName` calls `client.Store.Contacts.GetContact()` to resolve the contact name. If the WhatsApp server has not yet synced the contacts from the phone, `GetContact` returns an empty `FullName` — and the bridge falls through to `jid.User` (the raw phone number) as the chat name.

This happens even when the contact **is** saved in the local `whatsapp.db` → `whatsmeow_contacts` table, which is populated by whatsmeow from the last successful sync.

## Solution

Add a fallback that queries the local `whatsmeow_contacts` SQLite table when the server-side lookup fails or returns an empty name:

```go
SELECT COALESCE(NULLIF(full_name,), NULLIF(push_name,), NULLIF(first_name,))
FROM whatsmeow_contacts WHERE their_jid = ?
```

The name resolution order becomes:

1. `client.Store.Contacts.GetContact()` — server-side (unchanged)
2. **NEW:** local `whatsmeow_contacts` table — client-side fallback
3. `sender` parameter — existing fallback
4. `jid.User` — last resort (phone number)

## Changes

| File | Change |
|---|---|
| `whatsapp-bridge/main.go` | Add `waDB *sql.DB` to `MessageStore` |
| `whatsapp-bridge/main.go` | Open `store/whatsapp.db` in `NewMessageStore()` |
| `whatsapp-bridge/main.go` | Close `waDB` in `Close()` |
| `whatsapp-bridge/main.go` | Add `whatsmeow_contacts` fallback in `GetChatName()` |

## Testing

- `go test ./...` — **99/99 pass**
- The fallback is guarded by `messageStore.waDB != nil`, so existing tests (which use `newTestMessageStore` with an in-memory DB) continue to work unchanged
- Verified locally: 121 individual chats that showed phone numbers now show correct contact names

## References

The AGENTS.md "Gotchas" section already documents that `whatsapp.db` contains the contact data used by whatsmeow and recommends treating it as opaque — this PR only reads from it, never writes.


jack-arturo (approved):
Resolved the contact fallback review by scoping lookup to the active account, opening the WhatsMeow DB read-only/no-create, adding fallback tests, and verifying Go tests/build/lint locally.

PR #144, opened 2026-06-24
perf(bridge): index messages(chat_jid) to speed up LID migration

davidstrouk:
## Summary

- The `messages` table is keyed by `PRIMARY KEY (id, chat_jid)`, so queries filtering on `chat_jid` alone can't use that composite index and fall back to full table scans.
- The startup LID→phone migration (`MigrateLegacyLIDChatsToPhoneJIDs` / `MigrateLegacyLIDSendersToPhones`) joins and deletes by `chat_jid`, so on a large store every mapped chat scans the whole `messages` table — first-run startup hangs for minutes.
- Add `idx_messages_chat_jid` next to the existing `calls(chat_jid)` index.

## Type of change

- [ ] `fix` — bug fix
- [ ] `feat` — new feature
- [x] `chore` / `docs` / `ci` / `refactor` / `test` / `perf`
- [ ] Breaking change (`!` in commit, or `BREAKING CHANGE:` in body)

## Scope check

- [x] This change fits the [`ROADMAP.md`](https://github.com/verygoodplugins/whatsapp-mcp/blob/main/ROADMAP.md) "in scope" list, **or** I've opened an issue first to discuss

## Testing

Reproduced on a real store (~100k messages / ~43 MB) migrated from a legacy `@lid`-keyed bridge: the first-run LID migration stalled for several minutes (full scans of `messages`), pegging startup before the transaction committed. After adding the index, the same migration completes and the bridge connects in ~1 s. Subsequent startups are unaffected (migration is a no-op once there are no `@lid` rows).

- [x] Added or updated tests — `TestNewMessageStoreCreatesMessagesChatJIDIndex` asserts the index is created by `NewMessageStore()`
- [ ] Ran `uv run pytest -v` (Python changes) — n/a, Go only
- [x] Ran `golangci-lint run` and `go build ./...` (Go changes) — `go build ./...` and `go test ./...` pass; `gofmt` clean (golangci-lint will run in CI)

## Docs

- [ ] Updated `README.md` (if user-visible) — n/a (internal schema/perf only)
- [ ] Updated tool descriptions in `whatsapp-mcp-server/main.py` (if MCP tools changed)
- [ ] Updated `.env.example` (if env vars changed)

## Risk / rollback

Minimal and additive. `CREATE INDEX IF NOT EXISTS` is a no-op on databases that already have the index; it only adds an index (no data/schema rewrite). Rollback: drop the index or revert the one-line change.


jack-arturo (approved):
Verified against current main: Go tests, build, and golangci-lint pass locally after refreshing the branch.

PR #149, opened 2026-07-01
fix(bridge): preserve original timestamp on retry-redelivered messages

malyszm:
## Summary

- Incoming messages that fail to decrypt on first delivery (common after the bridge has been offline a while — an outage, or the periodic linked-device session expiry, when the local Signal session lacks the current sender key) are re-sent by WhatsApp after a retry receipt. The re-sent stanza's `t` is the **resend time**, not the original send time, so those messages were stored at reconnect-time — corrupting recency-ordered views (last-message-per-chat and chat sort order) until they aged out.
- Fix: cache the original timestamp from `events.UndecryptableMessage` (the first delivery carries the true `t`) and reuse it when the decrypted retry lands in `handleMessage`. Guarded to only override when strictly earlier; entries consumed on use with a soft size cap. Applied to the content and reaction store paths and the chat's last-message time.

## Type of change

- [x] `fix` — bug fix

## Scope check

- [x] This change fits the `ROADMAP.md` "in scope" list (reliability fix for the Go bridge — reconnect/session-recovery correctness)
- [x] PR is focused on one concern
- [x] PR is ≤ ~300 LOC (~140)

## Testing

- [x] Added or updated tests — unit tests for the timestamp-cache helpers (earliest-wins, consume-on-take, empty/zero guards)
- [x] Ran `golangci-lint run` and `go build ./...` (Go changes) — `go build ./...`, `go vet ./...`, `gofmt`, and `go test ./...` all clean locally (golangci-lint deferred to CI)
- [x] Manually exercised the affected code path — reproduced after a multi-week offline gap: several incoming messages that failed initial decryption were stored with the reconnect time; with this change the cached original timestamp is used instead (the bridge logs `Using original pre-retry timestamp for <id>…` when it fires)

## Docs

- No user-visible surface or env/tool/schema change — internal timestamp-correctness fix, no docs update needed.

## Risk / rollback

- Low. The cache only ever *lowers* a message's timestamp toward its true send time, and only when an earlier value was observed for that exact message ID; otherwise behavior is unchanged. Memory is bounded (consumed on use + soft cap). Revert is a single commit.

### DIFF
--- a/whatsapp-bridge/main.go
+++ b/whatsapp-bridge/main.go
@@ -17,6 +17,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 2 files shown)


jack-arturo:
@codex can you give this a quick look before we merge?

jack-arturo (dismissed):

malyszm:
Addressed the Codex feedback in a094d4f — extractMediaInfo now gets the same retry-corrected timestamp that gets stored, so MediaFilename matches the on-disk name downloadMedia rebuilds. @jack-arturo should be good to merge now.

jack-arturo (approved):

PR #155, opened 2026-07-09
feat(bridge): persist chat read state from read receipts + history-sync backfill

### LINKED ISSUE #194: feat: persist chat read state in messages.db (chats.last_read_time)
### Problem

`messages.db` has no notion of read state, so an MCP client can only approximate "unread" as *the latest message in this chat is inbound*. That heuristic surfaces long-dead one-sided threads and every chat already read on the phone or another linked device — the two cases an agent most needs to tell apart.

### Use case

"Which chats are still waiting on me?" — the most common triage question asked of the MCP server. Today `list_chats` gives an agent no way to answer it: it either replays chats the user read on their phone hours ago, or misses genuinely unread ones sitting under an old outbound message.

### Scope bucket

MCP correctness.

### Proposed approach (schema change — `chats.last_read_time`)

Per `CONTRIBUTING.md`, schema changes to `messages.db` need discussion before a PR lands, so filing this to cover the one in #155.

**Bridge (write path)**

- Add `chats.last_read_time TIMESTAMP` via the existing `ensureColumn` migration — additive, existing stores upgrade in place, nothing reads it unless present.
- Populate from two sources:
  1. **Read receipts** (going forward) — an `*events.Receipt` handler records *self*-reads (`read-self` for DMs, `read` with `IsFromMe` for groups), resolving the receipt's `@lid` chat to the phone JID so the marker key matches the message rows.
  2. **History-sync backfill** (existing chats) — a conversation WhatsApp reports as fully read (`unreadCount == 0 && !markedAsUnread`) is marked read up to its last message.
- The marker merges **monotonically**, mirroring the existing `last_message_time` merge in `StoreChat`, so out-of-order receipts and late history-sync chunks can never move it backwards and un-read a chat.

**MCP (read path)**

- Every chat tool (`list_chats`, `get_chat`, `get_direct_chat_by_contact`, `get_contact_chats`) gains two fields: `last_read_time` (the marker) and `unread` (derived: the last message is inbound and newer than the marker).
- Stores written by an older bridge lack the column; the read path detects that and reports `last_read_time: null`, degrading `unread` to the old "last message is inbound" heuristic rather than erroring.

### Known limitation

Because the marker only advances, manually re-marking an already-read chat as *unread* on the phone isn't reflected. It's uncommon, and monotonicity is what keeps the marker idempotent and race-safe against out-of-order receipts. A separate `marked_unread` column could cover it later if it turns out to matter.

### Alternatives considered

- **Keep the heuristic client-side.** No schema change, but it's wrong in exactly the cases that matter (already read elsewhere / last message is old and outbound), and every consumer reimplements it.
- **Store WhatsApp's unread *count*.** Richer, but only arrives via history sync — receipts don't carry it, so it would go stale immediately after the initial sync. A monotonic timestamp is the piece both sources can actually keep current.

sasha-id:
Closes #194

## What & why

The bridge's `messages.db` has no notion of read state, so consumers can only approximate "unread" as *the latest message in a chat is inbound* — which surfaces long-dead one-sided threads and anything already read on another device. This adds a `chats.last_read_time` marker and exposes it (plus a derived `unread` flag) through the MCP chat tools, so clients can distinguish genuine unread from "latest message is inbound".

### Bridge — write path

Two sources populate a single monotonic `chats.last_read_time`:

1. **Read receipts (going forward).** A new `*events.Receipt` handler records *self*-reads — DM reads arrive as `read-self`, group reads as `read` with `IsFromMe` (the participant is us). It resolves the receipt's `@lid` chat to the phone JID via the existing `resolveLIDChat` helper (receipts carry no phone-JID alt, so this leans on its `LIDs.GetPNForLID` fallback), so the marker key matches the message rows.

2. **History-sync backfill (existing chats).** In the existing `handleHistorySync` conversation loop, a conversation WhatsApp reports as fully read (`GetUnreadCount() == 0 && !GetMarkedAsUnread()`) is marked read up to its last message.

`MarkChatRead` merges the marker **monotonically** — mirroring `StoreChat`'s existing `last_message_time` merge — so out-of-order receipts and late history-sync chunks can never move it backwards and un-read a chat. The column is added through the existing `ensureColumn` migration, so existing stores upgrade in place.

### MCP — read path

Every chat tool (`list_chats`, `get_chat`, `get_direct_chat_by_contact`, `get_contact_chats`) returns two new fields:

```jsonc
{
  // …existing fields…
  "last_read_time": "2024-01-15T09:00:00+00:00", // the marker, null if never reported
  "unread": true                                 // last message is inbound and newer than the marker
}
```

`unread` is derived on the Python side (`Chat.unread`): a chat whose last message is ours is never unread, and a chat with no marker degrades to the old "last message is inbound" heuristic. To make that work in both modes, `list_chats`/`get_chat` now always join the last message for `is_from_me`; `include_last_message` still gates its content/sender.

**Backward compatibility:** a `messages.db` written by a bridge older than the migration has no `last_read_time` column. The read path detects that (`PRAGMA table_info`) and selects a NULL literal instead, so an un-upgraded store keeps working and simply reports `last_read_time: null`.

Documented in `README.md` (chat response shape + a "Read state" section with the caveats) and in the `list_chats`/`get_chat` tool docstrings.

## Known limitation

Because the marker only advances, manually marking an already-read chat as *unread* on the phone isn't reflected. It's uncommon, and monotonicity is what keeps the marker idempotent and race-safe. A future `marked_unread` column could add it.

## Tests

- **Go** (`main_test.go`): the monotonic merge (`MarkChatRead` sets when absent, advances on a newer read, never regresses on an older one) and the self-read classification (`isSelfReadReceipt` across read-self / group-read-by-us / other-user-read-our-message / delivered / sender).
- **Python** (`tests/test_read_state.py`): end-to-end read path over a real SQLite store — `last_read_time` and `unread` across all four chat tools and both `include_last_message` modes, covering unread, read-elsewhere, last-message-is-ours, and no-marker chats, plus a store with no `last_read_time` column at all.

Full bridge suite, `go vet`, `go build`, `pytest` (76 passed), and `ruff check`/`format` pass.

## Notes

Verified against a live linked account that read receipts arrive and resolve to the correct phone JID.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,6 +272,42 @@ Download media from a received message.
 
 ### Chat Operations
 
+All chat tools (`list_chats`, `get_chat`, `get_direct_chat_by_contact`,
+`get_contact_chats`) return the same chat shape:
+
+```jsonc
+{
+  "jid": "1234567890@s.whatsapp.net",
+  "name": "Alice",
+  "is_group": false,
+  "last_message_time": "2024-01-15T10:30:00+00:00",
+  "last_message": "hello world",       // null when include_last_message=false
+  "last_sender": "1234567890",         // null when include_last_message=false
+  "last_is_from_me": false,
+  "last_read_time": "2024-01-15T09:00:00+00:00", // how far the chat is read
+  "unread": true                       // last message is inbound and unread
+}
+```
+
+#### Read state (`last_read_time` / `unread`)
+
+`last_read_time` is the bridge's read marker for the chat, fed by read
+receipts from your own devices and backfilled from history sync. `unread` is
+derived from it: true when the chat's last message is inbound and newer than
+the marker. This distinguishes a genuinely unread chat from one whose last
+message merely happens to be inbound but was already read on the phone.
+
+Caveats:
+
+- **The marker only moves forward.** Marking an already-read chat as *unread*
+  again on the phone is not reflected.
+- **No marker means no read was ever reported** — for a chat with an inbound
+  last message, `unread` then falls back to the old heuristic and reports
+  true. Stores written by a bridge older than the `chats.last_read_time`
+  column report `last_read_time: null` and behave the same way.
+- **`unread` is a chat-level flag, not an unread count.** WhatsApp's unread
+  counter is not persisted.
+
 #### `list_chats`
 
 List all chats with metadata.
--- a/whatsapp-bridge/main.go
+++ b/whatsapp-bridge/main.go
@@ -196,6 +196,9 @@ func ensureMessageStoreSchema(db *sql.DB) error {
 	if err := ensureColumn(db, "chats", "ephemeral_setting_timestamp", "INTEGER NOT NULL DEFAULT 0"); err != nil {
 		return fmt.Errorf("failed to ensure chats.ephemeral_setting_timestamp column: %w", err)
 	}
+	if err := ensureColumn(db, "chats", "last_read_time", "TIMESTAMP"); err != nil {
+		return fmt.Errorf("failed to ensure chats.last_read_time column: %w", err)
+	}
 	if err := ensureColumn(db, "messages", "deleted_at", "TIMESTAMP"); err != nil {
 		return fmt.Errorf("failed to ensure messages.deleted_at column: %w", err)
 	}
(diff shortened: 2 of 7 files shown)


jack-arturo (dismissed):
Please redesign this as a user-visible feature before merge. `last_read_time` expands the schema but is not exposed through any MCP chat/read output, so clients cannot use the stored state. Open the required schema-change issue, document the MCP response shape, and add an end-to-end read-path test.

sasha-id:
Addressed in bae53d6.

- **Exposed through MCP.** `last_read_time` and a derived `unread` flag are now returned by `list_chats`, `get_chat`, `get_direct_chat_by_contact` and `get_contact_chats`. `unread` is true only when the last message is inbound *and* newer than the marker, so a chat read on the phone no longer reports as unread. `list_chats`/`get_chat` always join the last message now (needed for `is_from_me`); `include_last_message` still gates its content.
- **Schema-change issue:** #194, with the write path, read path, and the alternatives considered. PR now `Closes #194`.
- **Response shape documented** in README (chat shape + a "Read state" section with the caveats: forward-only marker, no-marker fallback, no unread count) and in the `list_chats`/`get_chat` tool docstrings.
- **End-to-end read-path test:** `tests/test_read_state.py` drives all four tools against a real SQLite store across unread / read-elsewhere / last-message-is-ours / no-marker chats and both `include_last_message` modes — plus a store with no `last_read_time` column, which the read path now handles by selecting a NULL literal instead of erroring (an un-upgraded bridge store keeps working).

jack-arturo (approved):

jack-arturo:
Thanks @sasha-id looking really good 🙏. Will merge pending a quick @codex review.

jack-arturo (approved):
Re-reviewed after bae53d6 — MCP exposure, schema issue #194, docs, and `tests/test_read_state.py` address the earlier request.

Merging as-is; Codex P2s (LID migration `last_read_time`, history-sync field presence, deterministic last-message join, None-as-unknown, receipt message-boundary) land in a same-day follow-up with the #200 polish.

jack-arturo:
Merged. Codex P2 polish + mark-read→local `last_read_time` glue: #203.

PR #158, opened 2026-07-13
fix(bridge): refuse to start when another bridge holds the session

wu-liang:
## Problem

Two bridges sharing a store share one WhatsApp device session, and WhatsApp permits a single live connection per device. The newcomer's login evicts the incumbent with `<stream:error><conflict type="replaced"/>`; the incumbent reconnects and evicts the newcomer; the two flap indefinitely.

While they flap, **neither reliably persists incoming messages** — yet `/api/health` on whichever socket survives still reports `connected:true`. Message sync therefore stops with no error surfaced anywhere: the message table simply stops growing, which is indistinguishable from a quiet day. I lost a fair while to exactly this before finding the conflict frames in the bridge log:

```text
[Client WARN] Received stream end frame
[Client INFO] Got replaced stream error, sending StreamReplaced event
<stream:error><conflict type="replaced"/></stream:error>
```

It is easy to hit by accident: `launchctl`/systemd already has a bridge running, you open a terminal and run `./whatsapp-bridge` to check something, and sync quietly dies.

## Root cause

The REST port was the only thing standing in a duplicate's way, and it was claimed too late to be of any use:

- `startRESTServer` ran **after** `client.Connect()`. By the time a port clash was discoverable, the duplicate had already evicted the running bridge from the session.
- A failed `ListenAndServe` was merely printed from inside a goroutine, so the duplicate kept running regardless — attached to WhatsApp, serving nothing.

## Fix

Claim the store lock and the REST port **before** opening the session database or dialling WhatsApp, and exit if either is taken:

- exclusive `flock` on `store/.bridge.lock`, recording the holder's PID so a refused start can name the incumbent;
- `net.Listen` on the REST port up front — `startRESTServer` now serves on the pre-bound listener.

The flock is held by the open file description, so the kernel drops it when the process dies: a crashed bridge never locks its own restart out (covered by a test).

## Security-sensitive: network bind

Flagging this explicitly per PR rule 8 in `AGENTS.md`, since it moves where the REST listener is bound:

- **Still loopback-only.** `net.Listen("tcp", "127.0.0.1:<port>")` — the same address as before, so there is no new exposure. A test asserts the bound address is a loopback address.
- **Bind failure is now fatal** rather than ignored. Previously a bridge whose port was taken would run on with no REST API at all; now it exits.
- **No new dependencies** — stdlib `net`, `syscall`, `path/filepath` only.

## Behaviour change

The bridge now exits non-zero, instead of starting, when another bridge holds `store/.bridge.lock` or the REST port is already in use. Documented in `README.md` (Troubleshooting → "Bridge Won't Start") and `AGENTS.md` (Gotchas → "One bridge per store").

## Tests

New `whatsapp-bridge/singleton_test.go`:

- `TestResolveBridgePort` — default / valid override / non-numeric / out-of-range / zero
- `TestAcquireBridgeLockRefusesSecondHolder` — a duplicate is refused
- `TestAcquireBridgeLockReleasedOnClose` — a crashed bridge does not lock out its restart
- `TestBindBridgePortRefusesPortInUse`
- `TestBindBridgePortBindsLoopbackOnly`

Locally green: `golangci-lint run ./...` (0 issues), `go vet ./...`, `go test ./...`, `gofmt -l`.

Verified end-to-end on macOS. With a bridge running under `launchd`, starting a second one now exits immediately:

```text
[Client ERROR] another whatsapp-bridge already holds store/.bridge.lock (pid 18771) —
refusing to start, because a second bridge would kick the running one off the WhatsApp session
```

and the incumbent's `stream replaced` count stays flat — it is never evicted, and message sync continues uninterrupted.


jack-arturo (changes requested):
The reliability objective is worthwhile, but this needs a cross-platform implementation before merge:

- `syscall.Flock` does not compile on Windows. Isolate it with build tags plus an equivalent Windows strategy, or use a portable lock.
- Add a `GOOS=windows` compile gate.
- Rebase the branch onto current `main` once the portability fix is in.

PR #163, opened 2026-07-16
feat: add Windows live pairing page (auto-refreshing QR in browser)

Muhfaden:
## What

Adds `scripts/live-pairing/` — a Windows pairing flow that shows the QR code in an auto-refreshing browser page instead of the terminal:

- `pair-live.ps1` — runs `whatsapp-bridge.exe` hidden with stdout redirected to a log, picks up each rotating pairing payload from whatsmeow's `Emitting QR code <payload>` QRChannel debug line, renders it to SVG, and rewrites an auto-refreshing `out/pair.html`. Flips to a success page on `Successfully connected and authenticated!` and exits, leaving the bridge running. Restarts the bridge when a pairing window expires; gives up after 30 minutes.
- `payload2svg.py` — converts the latest logged payload to a compact SVG QR (needs `pip install qrcode`).
- `pair-whatsapp.cmd` — double-click launcher (starts the watcher hidden, opens the page).
- `README.md`, plus `out/` is gitignored.

## Why

Pairing QR codes rotate every 20–60 seconds. On Windows the terminal QR frequently renders as mojibake (default codepage/raster fonts), and the bridge only prints the ASCII QR once per connection attempt (`qrCodeShown` in `main.go`), so the printed code goes stale while fresh codes exist only in debug logs. Users scan a stale code and the phone shows "Can't link device" with no error on the bridge side. A browser page that always shows the current code removes the race entirely.

## Notes for reviewers

- Windows-only by design (that's where the terminal QR pain is worst); the Python SVG converter is platform-neutral if anyone wants to build a macOS/Linux equivalent.
- The watcher reads the redirected log with `FileShare.ReadWrite` — `Start-Process -RedirectStandardOutput` holds the file with a write lock, and a plain `ReadAllText` fails on it (this cost me a debugging round).
- It parses the `Emitting QR code` line from whatsmeow's DEBUG logger, so it depends on the bridge keeping the DEBUG log level for the client; an alternative would be patching `main.go` to expose codes explicitly, which I kept out of scope to leave the bridge untouched.
- Tested end-to-end on Windows 10 with Python 3.12 and a CGO/WinLibs-built bridge.


jack-arturo:
Closing: this flow relies on parsing DEBUG logs, stores pairing payloads in local generated files, introduces an undeclared Python dependency, and uses `taskkill /im whatsapp-bridge.exe /f`, which can terminate unrelated bridge instances. A future pairing improvement should use whatsmeow's supported PairPhone primitive instead.

PR #166, opened 2026-07-18
delete_message tool, optional caption on send_file, and a media-download 403 fix

vyakunin:
Three independent commits; happy to split into separate PRs.

### 1. `delete_message`

Adds a delete/revoke tool. `for_everyone: true` revokes for both sides ("Delete for everyone", only valid for messages this client sent); `false` removes it locally so `list_messages` stops returning it while the other party still sees it. Takes `chat_jid` + the whatsmeow `message_id` returned by `send_message`.

### 2. `send_file` caption

`send_file` could only send a bare attachment. Adds an optional `caption` rendered with the file, matching what the app does. Omitting it preserves current behaviour.

### 3. Media-download 403

Downloads were failing with 403 on some media. The signed query string was being stripped from `directPath` before the fetch, so the request arrived unsigned. Keeping the query intact fixes it.

Each commit stands alone and can be cherry-picked.


jack-arturo:
Closing this combined PR because it mixes delete-message, attachment captions, a media fix already merged in #132, Cyrillic transliteration, and a broad whatsmeow upgrade. Please resubmit any desired caption or delete-message work as focused, current-main PRs with authenticated bridge calls and tests.

Issue #36, opened 2026-04-18
feat: CLI flag to request full history at pair time


edmenendez:
## Problem

whatsmeow's default `DeviceProps.RequireFullSync = false` asks the primary device for "recent" history only at pair time — in practice, ~3 months, with the phone deciding the exact window. There's currently no way to request more history via the bridge. Users who want to pull older messages on a fresh pair have no affordance.

On-demand history sync (via `BuildHistorySyncRequest` / `ON_DEMAND` events) is an alternative in theory, but community experience (see `tulir/whatsmeow#759`, `#654`, `#33`) shows WhatsApp's primary device typically drops those requests silently — confirmed locally on three test chats, zero messages ever returned. So the only practical lever for pulling more history is the pair-time configuration.

## Proposed solution

Add a CLI flag `--full-history-pair` to the Go bridge that overrides `store.DeviceProps` before client creation when passed:

```go
store.DeviceProps.RequireFullSync = proto.Bool(true)
store.DeviceProps.HistorySyncConfig = &waCompanionReg.DeviceProps_HistorySyncConfig{
    FullSyncDaysLimit:   proto.Uint32(3650),    // 10 years
    FullSyncSizeMbLimit: proto.Uint32(102400),  // 100 GB
    StorageQuotaMb:      proto.Uint32(102400),
}
```

- **Defaults to off.** No behavior change for normal launchd/systemd/Docker-managed restarts.
- **Only effective on a fresh pair** (no existing `whatsapp.db`). No-op for already-paired sessions because no pair handshake fires.
- **Non-destructive.** `messages.db` preserved across the re-pair; existing rows upserted identically.

## Measured result

Tested on a live install (iOS WhatsApp, 2026-04-18):

| Metric | Before | After |
|---|---|---|
| Total messages | 4,441 | **52,637** |
| Chats | 423 | 423 |
| Oldest on most active group | 2026-01-20 | **2023-10-25** |
| Oldest on other active chats | 2026-01-15 | **2023-04-20** |

**+48,196 messages pulled**, covering ~3 years of history that was previously unreachable. Initial sync took ~10 minutes wall time. No duplicates (bridge's existing `INSERT OR REPLACE` on `(id, chat_jid)` handles the overlap).

## Caveats to document

- The phone decides the actual cap. 10-year request ≠ 10-year result. iPad companion is documented at ~1 year; iOS primary's policy for other linked devices is not officially documented.
[...]

PR #168, opened 2026-07-20
feat(bridge): add on-demand history sync for a single chat

### LINKED ISSUE #167: feat: on-demand history sync - request older messages for a specific chat at runtime
### Problem

After pairing, the bridge can only store the history the **phone chooses to push**. There is no way to ask for older messages in a specific chat *later*, at runtime.

When the pair-time push sync delivers little or nothing (which varies a lot by phone, OS and account), the local `messages.db` is left with large gaps, and there is no recovery path short of destroying the session:

- Delete `whatsapp.db`
- Re-pair with `--full-history-pair`
- Re-scan the QR
- Re-sync everything, all-or-nothing

That is a heavy, lossy operation to fix what is often a single-chat gap.

On my account (Android primary, fresh pair, `--full-history-pair`), the pair-time push settled at **8 messages across 5 chats** and then plateaued — it stayed flat over a further 6 minutes of monitoring. A specific 1:1 chat I needed had exactly **1** message locally, despite years of real history on the phone.

### Use case

A user asks their MCP client something like *"summarise my thread with X"* or *"what did X say about the invoice last month?"*.

The agent calls `list_messages` / `get_contact_chats`, and that chat is empty or nearly empty locally — not because the history doesn't exist, but because the phone never pushed it. Today the only remedy is a full unpair/re-pair cycle, which is disruptive and still all-or-nothing.

What's missing is the ability to say: *"fetch older messages for **this one chat**, now."*

### Which scope bucket does this fit?

**Bridge stability.**

`ROADMAP.md` lists *"Pair-time configuration (e.g. history sync window) — additive, opt-in flags only"* as in scope, and #36 / #37 (`--full-history-pair`) established that controlling history sync belongs in this bridge.

This proposal is the **runtime complement** to that already-accepted pair-time control: same subsystem, same "the phone has the final word" caveat (`AGENTS.md` gotcha #4), but usable without destroying the session.

To be explicit about the delta: the roadmap bullet says *pair-time*, and this is *runtime*, so it is adjacent to that bucket rather than squarely inside it. Flagging that honestly rather than glossing over it — happy to be redirected if you see it differently.

### Proposed approach (optional)

whatsmeow already exposes the primitive: `Client.BuildHistorySyncRequest(lastKnownMessageInfo, count)`. Nothing needs to be reimplemented — it just isn't reachable today.

A minimal, additive endpoint:

```
POST /api/history
{ "chat_jid": "1234567890@s.whatsapp.net", "count": 50 }
```

- Anchors on the **oldest message already stored** for that chat, so the phone returns messages from before it.
- Sends the request as a peer message to the user's own JID (`SendRequestExtra{Peer: true}`).
- Results arrive through the **existing** `events.HistorySync` handler and are stored by the existing code path — no new storage or parsing logic.
- Opt-in: nothing changes for anyone who doesn't call it.

Caveat worth documenting: as with `--full-history-pair`, the phone ultimately decides how much it returns, so `count` is a request, not a guarantee.

I have this working locally against a real account — the endpoint pulled a 1:1 chat from 1 message to 44 on the first call.

### Alternatives considered

- **Re-pair with `--full-history-pair`** — works, but requires deleting the session, re-scanning the QR, and re-syncing everything. Far too heavy for a single-chat gap, and still phone-capped.
- **Waiting for the push sync to catch up** — doesn't help; in my testing it plateaued and stayed flat.
- **Querying the phone directly / another client** — out of the bridge's control and not automatable.
- **Downstream fork** — possible, but this is a small additive endpoint over an existing whatsmeow primitive, and the pair-time equivalent already lives here, so it seemed worth proposing upstream first.

### Pre-flight

- [x] I read `ROADMAP.md` and believe this fits the project's goals.
- [x] I searched existing issues and didn't find a duplicate.
- [x] I'm willing to help test or implement this (optional but appreciated).

Faiz-KraftedX:
## Summary

- Adds `POST /api/history`, which asks the phone for **older messages in a single chat at runtime**, anchored on the oldest message already stored for that chat.
- Fills the gap left by `--full-history-pair`: that flag only takes effect on a fresh pair, so recovering a gap in one chat otherwise means deleting `whatsapp.db`, re-scanning the QR and re-syncing everything.
- Wires up an existing whatsmeow primitive (`Client.BuildHistorySyncRequest`) — nothing is reimplemented. Results arrive through the **existing** `events.HistorySync` handler, so no new storage or parsing path is introduced.

## Type of change

- [ ] `fix` — bug fix
- [x] `feat` — new feature
- [ ] `chore` / `docs` / `ci` / `refactor` / `test` / `perf`
- [ ] Breaking change (`!` in commit, or `BREAKING CHANGE:` in body)

## Scope check

- [x] This change fits the `ROADMAP.md` "in scope" list, **or** I've opened an issue first to discuss — opened #167 first
- [ ] PR is ≤ ~300 LOC, **or** justified in the description

**LOC justification:** 383 insertions total, but only ~200 are new Go source — the rest is ~140 lines of tests and ~35 lines of docs. Issue #167 was opened first for exactly this discussion.

**Scope note (flagging honestly):** the roadmap bullet reads *"Pair-time configuration (e.g. history sync window)"*, and this is **runtime**, so it's adjacent to that bucket rather than squarely inside it. The argument for it belonging here is that #36/#37 already established history-sync control as in-scope for the bridge, and this is the runtime complement of that accepted feature. Happy to be redirected if you see it differently.

## Linked issues

Closes #167

## Testing

- [x] Added or updated tests
- [ ] Ran `uv run pytest -v` (Python changes) — **N/A, no Python changes**
- [x] Ran `golangci-lint run` (**0 issues**) and `go vet ./...` (clean)

**New tests** (`history_ondemand_test.go`, 18 subtests, all passing): `clampHistoryCount` bounds, `anchorTime` parsing across `time.Time` / `string` / `[]byte` / garbage, and `historyAnchorInfo` attribution for own-messages, bare-phone senders, full-JID senders, group chats and empty senders.

**Manual verification** against a real paired account: a 1:1 chat sitting at **1 stored message** went to **44** after a single `POST /api/history` call, with the messages landing in `messages.db` within seconds via the existing history-sync handler.

### Testing caveats — please read

Two things I want to be upfront about rather than have you discover them:

1. **I developed on Windows without a C compiler**, so everything local ran with `CGO_ENABLED=0`. Under that setting `go-sqlite3` compiles to its stub, so the pre-existing test `TestHandleHistorySync_LIDParticipant_ResolvedViaStore` fails locally with *"go-sqlite3 requires cgo to work. This is a stub."* I verified this failure is **not caused by this PR** by stashing my changes and reproducing it identically on a clean checkout of `main`. CI (which has cgo) should be unaffected, but worth a second look.
2. **The manual end-to-end run used a locally-swapped pure-Go SQLite driver** (again, no C compiler available). That swap is **not part of this PR** and nothing in this diff depends on it — the endpoint only touches `database/sql` and whatsmeow. Mentioning it so you can weight the manual evidence accordingly.

Caveat (1) is also why `anchorTime` accepts `time.Time`, `string` and `[]byte`: different SQLite drivers hand back the timestamp column differently, and the helper shouldn't be brittle about it.

## Docs

- [x] Updated `README.md` (if user-visible) — new "Requesting history for a single chat (on-demand)" section next to the pair-time docs, plus `/api/history` added to the REST API architecture diagram
- [x] Updated `AGENTS.md` / `CLAUDE.md` (if contributor-visible) — gotcha #4 now points at both the pair-time flag and the runtime endpoint
- [ ] Updated tool descriptions in `whatsapp-mcp-server/main.py` (if MCP tools changed) — **N/A, see note below**
- [ ] Updated `.env.example` (if env vars changed) — **N/A, no new env vars**

**Deliberately no MCP tool in this PR.** That would be a second concern and would pull in Python tests. There's precedent for bridge-only endpoints: `/api/typing` and `/api/health` both ship without a corresponding MCP tool. Happy to follow up with a separate PR exposing this as an MCP tool if you'd like it.

## Risk / rollback

- **Additive and opt-in.** No existing behaviour changes; callers that never hit `/api/history` see nothing different.
- **No new auth surface.** The endpoint goes through the existing `auth(...)` wrapper, so it inherits bearer-token auth and the loopback `Host` check like every other `/api/*` route.
- **No new outbound surface.** The only side effect is a peer message to the user's **own** JID (the history request itself). Nothing is sent to any third party, no files are read, and no new network binds are added.
- **Bounded input.** `count` is clamped to `[1, 500]` with a default of 50, so a typo can't request an unbounded backfill.
- **Rollback:** delete `whatsapp-bridge/history_ondemand.go` (+ its test) and the 3-line registration in `newRESTMux`. Nothing else references it.


Faiz-KraftedX:
Follow-up: I went back and audited how stable the whatsmeow primitive underneath this actually is, rather than assuming. Two things came out of it, one of which was a real fix — pushed in 722f089.

**The primitive itself looks like a reasonable thing to build on.** `BuildHistorySyncRequest` is a documented, exported method in `send.go`, sitting alongside `BuildEdit` / `BuildRevoke` / `BuildUnavailableMessageRequest`. It isn't `internal/`, isn't deprecated, and isn't flagged experimental. It maps onto a real WhatsApp peer-data-operation (`HISTORY_SYNC_ON_DEMAND`) — the same mechanism the official clients use for "load earlier messages" — so this isn't a hack around the protocol.

**The fix: I was using it at the wrong level.** whatsmeow's doc comment says the built message *"can be sent using `Client.SendPeerMessage`"*, but I had hand-rolled the equivalent:

```go
client.SendMessage(ctx, own, msg, whatsmeow.SendRequestExtra{Peer: true})
```

`SendPeerMessage` is exactly that call plus an `ErrNotLoggedIn` guard, resolving the own-JID internally via `getOwnID()`. Functionally identical today, but duplicating whatsmeow's addressing logic means it silently drifts if upstream ever changes how peer messages are addressed. Now switched to the documented call.

**Two other things I verified while in there:**

- `handleHistorySync` (`main.go:2848`) only *logs* `GetSyncType()` and then iterates `Data.Conversations` unconditionally — it doesn't filter by sync type. So `ON_DEMAND` responses are stored by the existing path with no changes needed. That's the code-level confirmation of the "no new storage path" claim in the PR description, rather than just the empirical result.
- whatsmeow documents 50 as the recommended count per request, so the default now cites that explicitly instead of looking like an arbitrary number.

**Residual risk, stated plainly:** whatsmeow publishes no tagged releases — this repo pins a pseudo-version (`v0.0.0-20260604205742-c6a4b703e48f`), so any of its APIs can change without a semver signal, and the underlying protocol is reverse-engineered (there's a field in this very request whose name says `MS` but which must carry seconds). That exposure is real, but it's the bridge's existing exposure across every whatsmeow call it already makes — this endpoint doesn't add a new *category* of risk, and after 722f089 it sits on the documented surface rather than a hand-assembled one.

whatsapp-bridge/history_ondemand.go
```diff
@@ -0,0 +1,213 @@
...
+		MessageSource: types.MessageSource{
+			Chat:     chat,
+			Sender:   senderJID,
+			IsFromMe: fromMe,
+			IsGroup:  chat.Server == types.GroupServer,
+		},
+		ID:        msgID,
+		Timestamp: ts,
+	}
+}
```
edmenendez: **Dead complexity: whatsmeow discards `Sender` and `IsGroup`, so most of this helper is inert.**

I checked `BuildHistorySyncRequest` at the exact commit pinned in `go.mod` (`c6a4b703e48f`). It reads only four fields of the `MessageInfo`:

```go
ChatJID:              proto.String(lastKnownMessageInfo.Chat.String()),
OldestMsgID:          proto.String(lastKnownMessageInfo.ID),
OldestMsgFromMe:      proto.Bool(lastKnownMessageInfo.IsFromMe),
OldestMsgTimestampMS: proto.Int64(lastKnownMessageInfo.Timestamp.Unix()),
```

`Sender` and `IsGroup` appear zero times in the function body. And it is not just the current implementation: the wire format itself (`HistorySyncOnDemandRequest` in `WAWebProtobufsE2E.proto`) has no field for the anchor's sender or a group flag, so no future whatsmeow version could start using them without WhatsApp changing the protocol.

That means the whole sender-resolution branch (bare-phone vs full-JID parsing, the `own` substitution, the chat fallback, the `IsGroup` flag) populates fields that are discarded before the request is built, and five of the `TestHistoryAnchorInfo` subtests exercise inert code. The doc comment above ("Sender matters for group chats") is unfortunately not true for this request type.

The helper can collapse to:

```go
info := types.MessageInfo{
    MessageSource: types.MessageSource{Chat: chatJID, IsFromMe: fromMe},
    ID:        msgID,
    Timestamp: ts,
}
```

which also lets `oldestStoredMessage` stop selecting `sender`. Worth keeping one test asserting the four load-bearing fields pass through correctly, since `IsFromMe` + `Timestamp` being right is what makes the phone accept the anchor - but the sender/group subtests currently give false confidence that that logic matters.

To be clear: nothing here is *wrong* at runtime (your 1 -> 44 manual result is consistent with this - the four fields that matter are all correct). It is purely simplification.

edmenendez (commented):
Nice PR - the core design is right: it wires an existing whatsmeow primitive to an existing storage path, and I verified that `handleHistorySync` stores `ON_DEMAND` syncs without filtering on sync type, so the "no new storage path" claim holds. Auth, input clamping, and logging all follow the conventions of the neighboring `/api/*` handlers.

One main finding inline (dead sender-resolution logic), plus two smaller notes:

1. `oldestStoredMessage` errors: the handler maps *every* DB error to 404 "no stored message". A locked or corrupt DB read would be reported as "chat is empty". Consider `errors.Is(err, sql.ErrNoRows)` for 404 and 500 for everything else.
2. Tests cover the pure helpers well, but `registerHistoryEndpoint` itself has no `httptest` coverage (method-not-allowed, missing `chat_jid`, not-connected, 404 anchor path). Cheap to add and they don't need cgo.

whatsapp-bridge/history_ondemand.go
```diff
@@ -0,0 +1,213 @@
...
+			writeErr(http.StatusBadRequest, fmt.Sprintf("Invalid chat_jid: %v", err))
+			return
+		}
+
+		id, sender, fromMe, ts, err := oldestStoredMessage(messageStore, req.ChatJID)
+		if err != nil {
+			writeErr(http.StatusNotFound,
+				"No stored message for this chat to anchor the request; send or receive one message first")
+			return
+		}
```
edmenendez: **Every DB error is reported as 404 "chat is empty", which misdirects the caller on real failures.**

`oldestStoredMessage` propagates any `Scan` error verbatim, and this branch turns all of them into 404 with "send or receive one message first". But `database/sql` only returns `sql.ErrNoRows` for an empty result set - a locked database (`SQLITE_BUSY` is easy to hit with the bridge's event handler writing on another connection), a corrupt page, or an I/O error arrives as a different error and would be answered with advice to go send a message, which cannot help.

Suggested split:

```go
id, sender, fromMe, ts, err := oldestStoredMessage(messageStore, req.ChatJID)
if errors.Is(err, sql.ErrNoRows) {
    writeErr(http.StatusNotFound,
        "No stored message for this chat to anchor the request; send or receive one message first")
    return
}
if err != nil {
    writeErr(http.StatusInternalServerError, fmt.Sprintf("Failed to look up anchor message: %v", err))
    return
}
```

(needs `database/sql` and `errors` added to the imports - the file currently has neither).

whatsapp-bridge/history_ondemand.go
```diff
@@ -0,0 +1,213 @@
...
+			"2006-01-02 15:04:05.999999999 -0700 MST",
+			time.RFC3339,
+		} {
+			if parsed, err := time.Parse(layout, t); err == nil {
+				return parsed
+			}
+		}
+	}
+	return time.Time{}
+}
```
edmenendez, line 91: **Nit (take it or leave it): the string layouts don't cover the one cross-driver case the fallback exists for.**

In production this branch is unreachable: cgo `go-sqlite3` intercepts columns *declared* `TIMESTAMP` on scan and hands back a `time.Time`, so the first case handles everything. The string fallbacks only run under a text-returning driver, i.e. the pure-Go driver swap you described in the PR body.

But the layouts cover the wrong text. I checked what each driver actually writes and ran the parses:

| Text in the DB (written by) | These 3 layouts parse it? |
|---|---|
| modernc default, Go `String()` format: `2026-08-01 14:00:00 +0000 UTC` | yes |
| RFC3339: `2026-08-01T14:00:00Z` | yes |
| cgo go-sqlite3 (`SQLiteTimestampFormats[0]`): `2026-08-01 14:00:00.123-06:00` | **no - all three fail** |

So a same-driver round trip (pure-Go writes, pure-Go reads - your local setup, which is why your manual test passed) works, but the scenario a fallback exists for - opening a real production `messages.db` written by cgo go-sqlite3 with the pure-Go dev build - hits the one format not in the list, `anchorTime` returns zero, and the endpoint 500s with "unreadable timestamp".

Impact is low (dev-only path, and the `ts.IsZero()` guard fails it cleanly), so feel free to punt. The cheap fix is to make the list unable to drift from the production driver by reusing its exported formats:

```go
import sqlite3 "github.com/mattn/go-sqlite3"

for _, layout := range append(
	[]string{"2006-01-02 15:04:05 -0700 MST", "2006-01-02 15:04:05.999999999 -0700 MST", time.RFC3339},
	sqlite3.SQLiteTimestampFormats...,
) {
```

(or just add `"2006-01-02 15:04:05.999999999-07:00"` and its `T` variant to the literal list, if importing the driver here feels heavy).

whatsapp-bridge/history_ondemand.go
Faiz-KraftedX: Confirmed and fixed in 4e99a7e. I checked the pinned commit independently: the `HistorySyncOnDemandRequest` proto message has no sender or group field at all (it's `ChatJID` / `OldestMsgID` / `OldestMsgFromMe` / `OnDemandMsgCount` / `OldestMsgTimestampMS` / `AccountLid` / `SupportInlineResponse`), and the builder reads only `Chat` / `ID` / `IsFromMe` / `Timestamp` — so the branch was inert at both levels, exactly as you said.

Collapsed `historyAnchorInfo` to those four fields, dropped `sender` from `oldestStoredMessage`'s SELECT and signature, removed the now-unused `own` JID in the handler, and replaced the five sender/group subtests with one that asserts the four load-bearing fields pass through. Also corrected the doc comment that wrongly claimed sender matters here. Good catch — thank you.

whatsapp-bridge/history_ondemand.go
Faiz-KraftedX: Fixed in 4e99a7e. Split it exactly as suggested:

```go
id, fromMe, ts, err := oldestStoredMessage(messageStore, req.ChatJID)
if errors.Is(err, sql.ErrNoRows) {
    writeErr(http.StatusNotFound, "No stored message for this chat to anchor the request; send or receive one message first")
    return
}
if err != nil {
    writeErr(http.StatusInternalServerError, fmt.Sprintf("Failed to look up anchor message: %v", err))
    return
}
```

Added `database/sql` and `errors` to the imports. Ordered the `sql.ErrNoRows` check before the generic `err != nil` so the sentinel isn't swallowed (it's a non-nil error too). Agreed on the `SQLITE_BUSY` concern given the event handler writes on a separate connection.

whatsapp-bridge/history_ondemand.go
Faiz-KraftedX: Fixed in 4e99a7e — took the literal route, with one caveat on the import option worth recording.

I verified your table against the driver source: cgo go-sqlite3 writes with `SQLiteTimestampFormats[0]` = `2006-01-02 15:04:05.999999999-07:00` and my three layouts miss it. But importing `SQLiteTimestampFormats` would be a problem here: it's defined in `sqlite3.go`, which is a cgo file (`import "C"`), so the identifier disappears under `CGO_ENABLED=0`. Referencing it would compile on CI but break any pure-Go build (including the dev setup this fallback exists to serve). So I added the two missing layouts as literals instead, with a comment explaining why not to swap them for the driver export:

```go
"2006-01-02 15:04:05.999999999-07:00",  // cgo go-sqlite3 write format
"2006-01-02T15:04:05.999999999-07:00",  // cgo go-sqlite3 write format, RFC3339-ish T
```

Added two `anchorTime` regression cases (space and T variants) that fail without the fix.

Faiz-KraftedX:
Thanks for the thorough review, @edmenendez — all three inline points and the two summary notes are addressed in 4e99a7e. Recap:

1. **Dead sender/group logic** → removed; helper collapsed to the four fields the request actually carries.
2. **Every DB error → 404** → split into `sql.ErrNoRows` (404) vs. other errors (500).
3. **Timestamp format gap** → added the cgo layouts as literals (not the driver export — it's cgo-gated and would break `CGO_ENABLED=0` builds).
4. **httptest coverage** → added `TestHistoryEndpointValidation`, covering method-not-allowed, invalid body, missing `chat_jid`, and not-connected. A nil client suffices there because `Client.IsConnected` is nil-safe, so no cgo is needed. The remaining paths (404-anchor and the success path) need a live paired client and a cgo-backed DB, so I've left those to manual/integration rather than fake the whole client surface — happy to add them if you'd prefer a stubbed client.

Net effect is a smaller, more honest surface: the endpoint now sets only load-bearing fields, reports failures accurately, and the tests no longer assert on discarded data. `golangci-lint run` is clean and `go test ./...` passes (minus the pre-existing cgo-only `TestNewMessageStore*` / `TestHandleHistorySync*` cases, which fail identically on `main` under `CGO_ENABLED=0`).

jack-arturo (approved):

PR #171, opened 2026-07-23
fix(bridge): read top-level participant for history-sync group senders

kramer65:
## Problem

After pairing, every group message imported via history sync is stored with the **group's own JID as its sender**, making it impossible to tell who wrote what in a group. Only `is_from_me` distinguishes the user's own messages.

On a freshly paired install, a sender breakdown of `messages.db` looked like this — every incoming group message attributed to the chat itself:

```
chat_jid                     sender                  count
120363162744727638@g.us      120363162744727638      998
31641798702-1426154712@g.us  31641798702-1426154712  1026
...
```

## Cause

Modern history syncs carry the group sender in the **top-level `WebMessageInfo.participant`** field, not in `Key.Participant`. The manual history-sync handling in `handleHistorySync` only read `Key.Participant` and fell back to the chat JID, so the fallback fired for every group message.

whatsmeow's own `ParseWebMessage` checks `webMsg.GetParticipant()` first and `Key.Participant` second — this PR mirrors that order.

## Fix

- `main.go`: read `msg.Message.GetParticipant()` first, fall back to `Key.GetParticipant()`. LID participants still resolve through `resolveUserJID` as before.
- `main_test.go`: regression test with the top-level participant set and `Key.Participant` empty. It fails on the previous code (sender = group JID) and passes with the fix.

Verified on a live install: after wiping the store and re-pairing with the fix, the same groups store real per-participant phone senders (349 distinct senders across ~5.5k group messages, zero rows with the group-JID-as-sender pattern).


jack-arturo:
@codex review

jack-arturo (changes requested):
Please restrict the top-level `Participant` fallback to group/broadcast-style history messages. On a one-to-one history item it can overwrite the actual sender with the recipient. Add a DM regression test; this is the unresolved review concern.

kramer65:
Addressed in 8a6cd92 — thanks, the concern was valid.

**Change:** the participant lookup (both top-level `WebMessageInfo.participant` and `Key.Participant`) is now only consulted when the chat server is *not* `s.whatsapp.net` / `lid` / `newsletter`; for those the sender is the chat peer itself. This mirrors the gate in whatsmeow's `ParseWebMessage` (`client.go:967`), and is a deny-list rather than a `g.us` allow-list so broadcast lists keep working the same way whatsmeow treats them. Note the pre-existing `Key.Participant` path had the same theoretical exposure, so the gate wraps both sources.

**Test:** `TestHandleHistorySync_DMSenderIgnoresParticipant` — a phone DM and a LID DM, each with a stray `Participant` set at both levels; asserts `sender == chat peer`. Verified it fails on the ungated code (`sender = 31699999999, want 31612345678`) and passes with the gate.

For context: on a freshly re-paired live install, 0 of 707 incoming DM history rows had `sender ≠ chat` — WhatsApp doesn't seem to set `participant` on DM rows today — but the gate is cheap and correct regardless.

Issue #178, opened 2026-07-30
Configurable, least-privilege access to WhatsApp conversations


jrenan:
## Problem

The WhatsApp MCP currently needs a safer way for operators to decide which conversations a model may access and what it may do in each one.

Access should not be all-or-nothing. Operators need to independently control whether the MCP can:

- read new messages;
- read previously stored history;
- send messages and actions;
- start a new individual conversation.

The configuration must remain local, use a credential that the MCP itself never receives, and be approachable for both Portuguese- and English-speaking operators.

Local deployments may also need to run the Bridge and its administration panel on a port other than the default.

## Expected behavior

- Unconfigured conversations are blocked by default.
- Each conversation has independent read-new, read-history, and send permissions.
- MCP tools and outgoing webhooks enforce the same access policy.
- A separate local admin credential protects permission changes.
- Operators can configure access through a responsive local panel.
- The panel supports Brazilian Portuguese and English.
- Translations are maintained in independent, testable catalogs.
- Starting new individual conversations is disabled by default.
- When explicitly enabled, a successful first message grants that new conversation full access.
- Unknown groups cannot be created through this permission.
- The Bridge port can be overridden for local deployment needs.
- Existing installations have a documented, temporary compatibility path during upgrades.

PR #179, opened 2026-07-30
feat: add configurable WhatsApp MCP access controls

jrenan:
## Summary

Adds a local, least-privilege access-control layer for the WhatsApp MCP.

Operators can now choose which conversations the model may access and independently allow reading new messages, reading stored history, and sending messages or actions. All unconfigured conversations are denied by default.

Refs #178

## What changed

- Added a persistent per-chat permissions schema to `messages.db`.
- Added independent permissions for:
  - reading new messages;
  - reading stored history;
  - sending messages and WhatsApp actions.
- Applied the policy to chat lists, searches, message reads, context retrieval, media downloads, reactions, typing indicators, and outbound sends.
- Applied the same read boundaries to outgoing message, media, and reaction webhooks.
- Added safe JID and LID resolution so permissions follow canonical conversations.
- Added migration behavior for existing LID-based permission rows.
- Added optimistic revisions to prevent one administration session from silently overwriting another.

## Local administration panel

The Bridge now serves a responsive permissions panel at `/admin/`.

The panel provides:

- conversation search and filtering;
- per-conversation permission switches;
- bulk permission presets;
- access summaries;
- confirmation dialogs for permission expansion;
- connected/disconnected Bridge status;
- responsive desktop and mobile layouts;
- accessible labels and live status messages.

The panel uses a dedicated admin token stored as `store/.admin-token`. This token is separate from `.bridge-token` and must never be exposed to the MCP server or model.

The admin endpoint is restricted to loopback access and includes session expiration, request-origin validation, login throttling, secure response headers, and constant-time token comparison.

## Languages

The entire administration experience is available in:

- Brazilian Portuguese (`pt-BR`);
- English (`en`).

The selected language is remembered locally in the browser. Translations are modularized in independent JSON catalogs under `whatsapp-bridge/admin/i18n/`.

Automated tests require both catalogs to contain the same complete set of keys and verify that every key referenced by the HTML and JavaScript exists.

## Starting new conversations

A disabled-by-default global permission can allow the model to start a new individual conversation.

When enabled:

- only a valid individual recipient may be used;
- unknown group JIDs remain blocked;
- quoted replies, reactions, and typing indicators cannot create a conversation;
- a successful first text or media message creates the conversation locally;
- the newly created conversation immediately receives full access: read new messages, read history, and send.

The global permission never overrides an explicit deny on an existing conversation.

## Compatibility

The updated MCP server fails closed when the permissions table is unavailable.

`WHATSAPP_MCP_LEGACY_ALLOW_ALL=true` is provided only as a temporary migration escape hatch for installations that must briefly run the new MCP server against an older Bridge.

## Security notes

This changes authentication and authorization behavior and should receive additional security review.

The permission system protects MCP tools, Bridge REST operations, and outgoing webhooks. It is not an operating-system sandbox: a process with direct access to `messages.db`, media files, or the admin token can bypass the application boundary.

Revoking access prevents future exposure but cannot retract data already delivered to a model.

## Documentation

Updated the README, environment example, and contributor environment-variable reference with:

- permission semantics;
- admin-token handling;
- new-conversation behavior;
- upgrade compatibility;
- webhook enforcement;
- filesystem-boundary limitations;
- bilingual panel usage.

## Validation

- `go test -race ./...`
- `go vet ./...`
- Go production build
- 98 Python tests
- `ruff check .`
- `ruff format --check .`
- JavaScript syntax validation
- translation-catalog parity tests
- desktop and mobile browser validation in both supported languages


jack-arturo:
Closing as out of scope. The 7,500-line access-control/admin-panel implementation introduces a local web UI and multi-user policy surface that conflicts with the roadmap's intentionally small, single-account bridge.

PR #180, opened 2026-07-30
feat: add a configurable Bridge port

### LINKED ISSUE #178: Configurable, least-privilege access to WhatsApp conversations

jrenan:
## Summary

Allows the WhatsApp Bridge to run on a port other than `8080`. The same listener also serves the local administration panel introduced by #179.

Closes #178

## What changed

- Added the `--port` command-line option.
- Kept `WHATSAPP_BRIDGE_PORT` as the environment-based configuration.
- Made the command-line value take precedence over the environment variable.
- Kept `8080` as the default when neither option is provided.
- Added validation for the valid TCP port range, `1-65535`.
- Added tests for defaults, environment configuration, command-line overrides, whitespace handling, and invalid values.
- Documented how to point `WHATSAPP_API_URL` at the selected Bridge port.

## Examples

```bash
cd whatsapp-bridge
go run . --port 9090
```

The MCP server can use the same Bridge instance with:

```bash
WHATSAPP_API_URL=http://localhost:9090/api uv run main.py
```

When combined with #179, the administration panel is available at:

```text
http://127.0.0.1:9090/admin/
```

## Validation

- `go test -race ./...`
- `go vet ./...`
- Go production build
- documentation review

### DIFF
--- a/README.md
+++ b/README.md
@@ -462,7 +462,21 @@ message DBs, media, and `.bridge-token`. Logs are left in
 
 | Flag                  | Default | Description                                                                                                                                                                                                                                                       |
 | --------------------- | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `--port`              | `WHATSAPP_BRIDGE_PORT` or `8080` | REST API port. The command-line value takes precedence over the environment variable. Valid range: `1-65535`. |
 | `--full-history-pair` | `false` | Request full history at pair time. Only takes effect on a fresh pair (no existing `whatsapp.db`); no-op for already-paired sessions. The phone ultimately decides the actual history window sent — see [Requesting full history](#requesting-full-history) below. |
+
+For a one-off launch on another port:
+
+```bash
+cd whatsapp-bridge
+go run . --port 9090
+```
+
+Point the MCP server at the same port:
+
+```bash
+WHATSAPP_API_URL=http://localhost:9090/api uv run main.py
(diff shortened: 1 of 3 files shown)


jack-arturo:
Closing this draft as unnecessary scope expansion: `WHATSAPP_BRIDGE_PORT` already provides configurable local deployment ports, and a second CLI setting is not needed for the current release plan.

PR #184, opened 2026-08-07
Docs: empty WEBHOOK_URL does not disable the webhook; ignore rotated logs

davidsimoes:
What

Two small corrections, no behaviour change.

1. AGENTS.md described `WEBHOOK_URL` as "empty = disabled". It is not. `webhook.go` falls back to `defaultWebhookURL`, so an operator who never sets the variable still gets every inbound message POSTed to `localhost:8769` for as long as the bridge runs. Both the AGENTS.md and README.md tables now say so, and say that turning the webhook off needs a code change rather than configuration.

2. `.gitignore` had `*.log`, which does not match logrotate output such as `bridge.log-20260803-113854.gz`. `bridge.log` is plaintext message history, so a rotated copy could become committable in a repo with a public remote. `*.log*` closes that.

Why

The docs claim was found by an operator who set `WEBHOOK_URL=""` expecting the webhook to stop, and instead saw the bridge keep POSTing to the compiled-in default: 10,466 consecutive connection-refused attempts on a long-running deployment. Nothing in the code is wrong here, only the documented contract.

The gitignore gap surfaced the first time log rotation was enabled on that same deployment.

Testing

Docs and gitignore only, no code paths touched. Verified against current `main` that `defaultWebhookURL` is still the fallback when the environment variable is unset, so the corrected text matches the code as it stands.

If you would rather have the behaviour changed than the docs (an early `return` on an empty `WEBHOOK_URL`), say so and I will open that instead. It is a breaking change for anyone relying on the default, which is why this PR only fixes the description.


jack-arturo (changes requested):
Both changes are reasonable, but please split them into independently reviewable PRs:

- a `docs:` webhook correction;
- a `fix:` rotated-log ignore change.

Please also use a conventional, lowercase title.

PR #186, opened 2026-08-08
fix: exit via os._exit in signal handler to avoid shutdown SIGABRT

antonio-mastropaolo:
## Problem

The MCP server aborts with **SIGABRT (Abort trap: 6)** on shutdown. macOS crash reports show:

```
handle_system_exit → Py_Exit → Py_FinalizeEx → flush_std_files
  → _io_TextIOWrapper_flush → buffered_flush → _enter_buffered_busy → abort()
```

i.e. CPython's fatal error:

> `_enter_buffered_busy: could not acquire lock for <_io.BufferedWriter name='<stdout>'> at interpreter shutdown, possibly due to daemon threads`

## Root cause

`shutdown_handler` (SIGINT/SIGTERM) called `sys.exit(0)`. That raises `SystemExit`, which runs **full interpreter finalization**. When a daemon thread spawned by the FastMCP/anyio stdio transport is mid-write to `stdout`/`stderr` at that instant, it holds the buffered-writer lock; finalization tries to flush/close the same stream, can't acquire the lock, and `abort()`s.

## Fix

Terminate from the signal handler with `os._exit(0)`, which skips interpreter finalization entirely, so the race cannot occur. Flush `stdout`/`stderr` first so buffered output isn't lost.

## Evidence

A stdout-lock stress harness (daemon thread hammering `stdout`, then SIGTERM):

| Handler | Result |
|---|---|
| old (`sys.exit`) | **60/60 SIGABRT (134)** |
| new (`os._exit`) | **0/60 aborts, 60/60 clean exit (0)** |

The real `main.shutdown_handler` was then verified directly: **25/25 clean, 0 aborts**.

## Tests / gates

- Adds `tests/test_shutdown_handler.py` (asserts `os._exit(0)`, that `sys.exit` is not used, streams flushed, and flush errors don't block exit).
- Full suite: **71 passed**.
- `ruff check` + `ruff format --check`: clean.

One concern, no dependency changes, no version/CHANGELOG edits.


jack-arturo (changes requested):
The `os._exit` direction matches the existing watchdog precedent, but this needs revision before merge:

- Do not flush stdout or stderr inside the signal handler; either flush can block on the same lock this change is intended to escape.
- Add a real subprocess regression test that holds the relevant writer/lock, sends SIGTERM, and asserts a bounded exit. The current monkeypatched `_exit` test proves control flow only.

PR #187, opened 2026-08-08
perf(bridge,mcp): reduce SQLite contention and unindexed scans in message context lookups

Edspan41:
## Summary
- messages.db defaulted to SQLite's rollback journal, so the bridge's continuous writes could block the MCP server's concurrent reads. Enable WAL mode + a `busy_timeout` on both sides to remove that contention.
- `get_message_context()` looked up the target message by `id` alone, but the `messages` table's primary key is `(id, chat_jid)`, so that query couldn't use an index and fell back to a full table scan. This runs once per row on every `list_messages()` call by default (`include_context=True`). Added `idx_messages_id` plus a composite `(chat_jid, timestamp)` index for the before/after context scans, and threaded `chat_jid` through `get_message_context()` so `list_messages()` uses the indexed lookup. The public `get_message_context` MCP tool signature in `main.py` is unchanged (chat_jid is an optional internal hint), so this is backward compatible.

## Why
Every `list_messages` call (the most common MCP read) was paying for a full-table-scan context lookup per returned row, and reads could contend with the bridge's writes under the default journal mode. Both were the dominant cost in the read path.

## Test plan
- [x] `go build` succeeds for the bridge
- [x] `uv run pytest` passes (72/72) in `whatsapp-mcp-server`
- [x] Verified indices and pragmas via `main.go` schema review


jack-arturo (changes requested):
Thanks—WAL and per-chat context lookups are worth pursuing, but this needs revision before merge:

- Remove `idx_messages_id`: the composite primary key `(id, chat_jid)` already indexes the `id` prefix, so this duplicates storage and write cost.
- Enable persistent WAL mode once from the bridge writer/migration path, not on every Python reader connection. `PRAGMA journal_mode=WAL` is a write-capable global mode change and can contend or fail.
- Add Python tests for the connection setup and for `list_messages → get_message_context(chat_jid)` behavior. This modifies Python code without corresponding coverage.

PR #195, opened 2026-08-09
feat(bridge): add WEBHOOK_ENABLED to opt out of outbound webhooks

edmenendez:
AGENTS.md documents WEBHOOK_URL as "(empty = disabled)", but an empty value falls back to defaultWebhookURL, so there is currently no way to turn outbound webhooks off. With no consumer listening on the default port, the bridge POSTs and logs a connection refused error for every inbound message.

This adds WEBHOOK_ENABLED (default true, read via the existing getEnvBool helper used by FORWARD_SELF) so operators can opt out without any change to current behavior.

* No behavior change unless WEBHOOK_ENABLED=false is set explicitly
* Leaves the implicit default URL and its token omission logic untouched, so TestSendWebhookOmitsBridgeTokenOnImplicitDefaultURL still passes
* Updates the AGENTS.md and README rows to describe actual behavior
* Adds a test covering the disabled path


jack-arturo (dismissed):

jack-arturo:
@codex review

jack-arturo:
Maintainer-owned successor #204 carries this rebased feature plus the three Codex review fixes. The contributor fork does not grant this maintainer credential write access to apply them here; please use #204 for the 0.6.0 release.

PR #198, opened 2026-08-11
feat(mcp): search messages through an FTS5 index

### LINKED ISSUE #197: Message search matches substrings, misses accented words, and scans the full table
### The problem

`list_messages(query=...)` filters with `instr()`, which is a raw substring test over `messages.content`. Three things follow from that, and they get worse the larger the database is.

**1. It matches inside words.** Searching for a short name or word returns every message where those characters appear anywhere. On my database, `ana` returns ~17k rows, almost all of them from `semana`, `analisado`, `imagina` and similar. The handful of messages actually about a person named Ana are buried.

**2. It is diacritic sensitive, so you have to guess the spelling.** People type accents inconsistently, and agents calling the tool have no way to know which spelling was used. On the same database:

| query | rows |
|---|---|
| `orçamento` | 271 |
| `orcamento` | 11 |
| `reunião` | 1419 |
| `reuniao` | 181 |

So typing the unaccented form silently loses ~96% of the matches. Nothing signals that anything was missed.

**3. Two words only match when they happen to be adjacent.** `boleto pagamento` is tested as one literal string, so a message containing both words a sentence apart does not match. There is no way to express "both of these words" or "either of these".

There is also a cost angle: every search is a full scan. On a 1.1M message database each query takes ~0.85s regardless of how selective it is, and that time is spent on every call an agent makes.

This feels like it belongs to the **"Search correctness across Unicode"** bucket under MCP correctness in `ROADMAP.md`. The same `instr()` path is what makes CJK behave the way it does today, so anything done here has to be careful not to regress scripts written without spaces between words.

### What I would expect

Searching for a word finds that word, regardless of accents, without also matching it inside unrelated words, and with some way to express more than one term. Ideally without adding a dependency or a manual migration step, and without breaking databases created by older versions.

### Environment

- MCP server: `whatsapp-mcp-server` at 0.5.1
- SQLite: 3.37.2 (FTS5 available)
- OS: Ubuntu 22.04 aarch64

I have a working implementation and tests for this and am happy to open a PR if the approach sounds right. Wanted to open the issue first per `AGENTS.md`.

filipemb:
Closes #197.

Replaces the `instr()` substring scan in `list_messages(query=...)` with an FTS5 index over `messages.content`.

### Behaviour change

| | before | after |
|---|---|---|
| `ana` | matches inside `semana`, `analisado` | matches the word `Ana` |
| `orcamento` | 11 rows | 259 rows, same as `orçamento` |
| `boleto pagamento` | only if adjacent | `AND`, `OR`, `NOT`, `"phrase"`, `prefix*` |
| ordering | date only | `sort_by="relevance"` ranks by BM25 |
| a query on 1.1M rows | ~0.85s full scan | under 10ms index lookup |

`sort_by="relevance"` is additive; `"newest"` and `"oldest"` are untouched, and a call without `query` builds exactly the SQL it built before.

### Design notes

**External content table.** `content='messages'` means the index stores only the inverted index, not a second copy of the text. On a 1.1M message database the file grew 13% (373 MB to 421 MB) and the initial build took 5.6s.

**Triggers, not application writes.** Three triggers keep the index current, so the Go bridge needs no change and does not need to know the index exists. The bridge writes with `INSERT ... ON CONFLICT(id, chat_jid) DO UPDATE`, which fires ordinary `INSERT` and `UPDATE` triggers. Worth stating explicitly because `INSERT OR REPLACE` would *not* have fired the delete trigger without `PRAGMA recursive_triggers`, and the sync would have rotted silently.

**Built on first search, not by a migration.** Existing databases pick it up with no extra step. If creation fails the error is printed once and the process falls back to the old scan permanently, so a read-only mount or a SQLite build without FTS5 degrades instead of breaking. The readiness check is cached per database path.

**CJK is deliberately left on the old path.** This was the one thing I got wrong on the first pass and the tests caught it. `unicode61` splits on non-word characters only, so `会議の資料を送ります` tokenizes as a *single* term and searching `会議` matches nothing, where `instr()` found it. Rather than regress those users, queries containing Hiragana, Katakana, Han or Thai keep using the substring scan. `trigram` would fix CJK but costs whole-word matching everywhere else, which is the main win here, so it did not seem like a good trade. Happy to revisit if you would rather have a tokenizer option.

**Untrusted search text cannot raise.** `"`, `(`, `-`, `*` and bare `AND`/`OR`/`NOT` are FTS5 operators, so a literal search like `obra (casa)` is a syntax error. On `OperationalError` the query is retried with every token quoted. Agents pass arbitrary user text into this parameter, so an exception here would be a routine failure mode.

### Verification

`tests/test_search.py`, 21 cases: diacritic folding, whole-word matching, multi-term operators, prefix, exact phrase, relevance sort, index auto-creation, trigger sync across insert/update/delete, the substring fallback, operator characters not raising, and script detection.

```
101 passed
ruff check .          All checks passed!
ruff format --check . 15 files already formatted
```

No new dependencies; `re` and `sqlite3` are stdlib.

### Notes for review

- `README.md` documents the query syntax, the automatic index build and both fallback cases.
- The tool docstring in `main.py` was updated so MCP clients see the syntax; that text is what the model actually reads when deciding how to call the tool.
- `list_chats(query=...)` still uses `instr()` over chat names. Chat counts are small and it is a different concern, so I left it alone.


filipemb:
Rebased onto current main (7f518e2) and pushed the follow-up fixes: FTS5 is now compiled in the bridge, CI/release paths, and the macOS installer, with regression coverage. Local gates are green: 120 pytest tests, Ruff check/format, uv lock check, Go tests/build with sqlite_fts5, version consistency, and the launchd installer tests. GitHub currently shows CI/Security as action_required because the fork workflows need maintainer approval. @jack-arturo, this should be ready for another look when you have a moment.

PR #200, opened 2026-08-11
feat: add explicit message read receipts

ingrqm:
## Summary

- Add an authenticated `/api/mark-read` bridge endpoint backed by `whatsmeow.Client.MarkRead`.
- Expose an explicit `mark_messages_read` MCP tool; listing or searching messages remains side-effect free.
- Validate message IDs, chat and sender JIDs, group participants, RFC 3339 timestamps, and bridge connectivity.

## Type of change

- [ ] `fix` — bug fix
- [x] `feat` — new feature
- [ ] `chore` / `docs` / `ci` / `refactor` / `test` / `perf`
- [ ] Breaking change (`!` in commit, or `BREAKING CHANGE:` in body)

## Scope check

- [x] This change fits the `ROADMAP.md` MCP correctness scope and was proposed in #199.
- [x] PR is focused on one concern.
- [x] PR is under ~300 changed lines.

## Linked issues

Closes #199.

This sends outbound receipts and is separate from the inbound read-state persistence in #194 / #155.

## Testing

- [x] Added or updated tests
- [x] `uv run ruff format --check .`
- [x] `uv run ruff check .`
- [x] `uv run pytest -v` — 80 passed
- [x] `go test ./...`
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] `golangci-lint v2.7.1` — 0 issues
- [ ] Manually exercised the external read-receipt side effect

## Docs

- [x] Updated `README.md`
- [x] Updated the tool description in `whatsapp-mcp-server/main.py`
- [x] No contributor instructions or environment variables changed

## Risk / rollback

The endpoint and MCP tool are additive and require an explicit call. Reverting the three commits removes the feature without changing either SQLite schema.

PR #201, opened 2026-08-11
feat: add explicit message read receipts

### LINKED ISSUE #199: feat: add explicit message read receipts
### Problem

Reading a message through the MCP server does not update its read state in WhatsApp. Messages handled by an agent therefore remain unread on the phone and other linked devices, and the MCP server currently has no explicit way to acknowledge them.

### Use case

After reviewing or handling specific incoming messages, a user wants to tell the agent to mark exactly those messages as read. This should be an explicit action rather than a side effect of `list_messages`, so merely searching or summarizing a conversation never sends a receipt.

### Scope bucket

MCP correctness.

### Proposed approach

Add an authenticated bridge endpoint backed by `whatsmeow.Client.MarkRead` and expose it as a `mark_messages_read` MCP tool.

The tool would accept:

- one or more message IDs from the same chat;
- the chat JID;
- the original sender JID for group messages, where WhatsApp requires the participant;
- an optional RFC 3339 timestamp, defaulting to the current time.

The bridge should validate JIDs and reject a group request without `sender_jid`. Receipts should only be sent when the tool is called explicitly.

This is separate from #194 / #155: those changes consume incoming receipt events to persist local chat read state, while this proposal sends an outbound receipt for selected messages.

### Alternatives considered

- Automatically mark messages read from `list_messages`: rejected because a read-only query should not create an external side effect.
- Leave the state unchanged: requires users to repeat the same work on their phone and leaves linked-device state inconsistent.

### Pre-flight

- [x] I read `ROADMAP.md` and believe this fits the project's MCP correctness goals.
- [x] I searched existing issues and did not find a duplicate.
- [x] I am willing to implement and test this.

jack-arturo:
## Summary

Maintainer merge vehicle for #200 with the Codex **P1** fix applied (couldn't push to the contributor fork).

- Add authenticated `/api/mark-read` + `mark_messages_read` MCP tool (from #200)
- Resolve PN→LID before `MarkRead` so migrated contacts actually receive receipts

Closes #199. Supersedes #200 (same commits + `0297f5f`).

## Test plan

- [x] `go test ./... -run 'MarkRead|ResolveRecipientJID_ForMarkRead'`
- [ ] CI green
- [ ] Close #200 after merge

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ A Model Context Protocol (MCP) server for WhatsApp, enabling Claude to read and 
 - **Message Management**: Search and read personal WhatsApp messages (text, images, videos, documents, audio)
 - **Contact Search**: Search contacts by name or phone number with `sender_display` format ("Name (phone)")
 - **Send Messages**: Send text messages to individuals or groups
+- **Read Receipts**: Explicitly mark selected messages as read across linked devices
 - **Media Support**: Send and download images, videos, documents, and voice messages
 - **Call History**: Capture incoming voice/video calls into a local SQLite table (live, 1:1 and group)
 - **Webhook Integration**: Forward incoming messages to external services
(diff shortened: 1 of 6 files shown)


jack-arturo:
Follow-up for remaining Codex P2s + local read-marker glue: #203.

PR #212, opened 2026-08-22
fix(bridge): resolve quoted-reply participant to a JID clients can match

Pchatain:
## Problem

When sending a reply (quoted message), `ContextInfo.Participant` was set to the raw quoted-sender value from the API request. The `messages.db` sender column stores bare user-parts, so replies quoted a participant like `15551234567` — no server, and in LID-addressed groups not even the right JID form. Recipient clients couldn't match it to any group member, so the quoted message rendered as **"You"** for every viewer and the quote notification misfired.

## Fix

Add `resolveQuotedParticipantJID`, which normalizes the value before it goes into `ContextInfo`:

- full JIDs (`@s.whatsapp.net`, `@lid`) are parsed and passed through
- bare numbers get the `s.whatsapp.net` server
- phone JIDs are upgraded to the cached LID mapping when available (`client.Store.LIDs.GetLIDForPN`), matching how LID-addressed groups identify their members

This mirrors the existing LID handling in `resolveMentionJIDs` / recipient resolution (see AGENTS.md gotcha #1 on JID-form mismatches).

## Testing

- New unit tests covering bare number → cached LID, bare number without LID, LID pass-through, phone JID → LID, and empty sender
- `go test ./...`, `go vet ./...`, and `go build` pass
- Verified manually: replies sent through the bridge now attribute the quoted message to the correct participant instead of "You"


Pchatain:
Generated with claude code, but I verified this fixed the problem in real usage.

PR #218, opened 2026-09-02
fix(bridge): document filename leaked the sender's absolute path

iB2:
### What's wrong

The document filename travels to the recipient inside the message, so every character of the path put there is published. The send path builds it with a manual split on the forward slash (`whatsapp-bridge/main.go`):

```go
Title:    proto.String(mediaPath[strings.LastIndex(mediaPath, "/")+1:]),
FileName: proto.String(mediaPath[strings.LastIndex(mediaPath, "/")+1:]),
```

On **Windows** that leaks the entire absolute path. The path reaching the send routine has already been through `filepath.EvalSymlinks` in the `/api/send` handler, which normalises separators to backslash — so `LastIndex(mediaPath, "/")` returns `-1`, the slice starts at index 0, and the recipient sees the whole path as the document's name:

```
C:\Users\<name>\.local\share\whatsapp-mcp\outbox\file.zip
```

Observed in the wild, not theoretically: a document sent to a group displayed the sender's home directory — username included — as the filename, to every recipient. Renaming the file does not help, because the directory is part of the leaked string.

Only **documents** are affected: video, image and audio messages carry no `FileName` field.

### The fix

`outboundFileName` in `media_path.go`. `filepath.Base` alone is not sufficient — it honours only the separator of the platform it runs on, so a Windows-style path handled by a Linux or macOS build still comes through whole. So it splits on **both** separators regardless of `GOOS`, drops a leftover drive prefix from paths like `C:file.zip` that have no separator at all, and falls back to a neutral constant rather than sending an empty filename.

### Tests

`TestOutboundFileNameNeverLeaksThePath`, table-driven over 7 cases: the reported Windows path, a path with spaces and parentheses, a POSIX path, mixed separators, an already-safe bare name, a drive-relative path with no separator, and a trailing separator. Beyond checking the expected name, each case asserts the real invariant — that nothing identifying the machine or the user (`Users`, `home`, `:`, `\`, `/`) survives into the name on any platform.

`go build ./...` and `go test ./...` pass. Three `TestMigrateLegacy*` tests fail on my machine both with and without this change — Windows file-lock on `t.TempDir()` cleanup, unrelated to this patch.

### Scope

One concern, 3 files, +106/-2.

### DIFF
--- a/whatsapp-bridge/main.go
+++ b/whatsapp-bridge/main.go
@@ -1444,8 +1444,12 @@ func sendWhatsAppMessage(client *whatsmeow.Client, messageStore *MessageStore, r
 			}
 		case whatsmeow.MediaDocument:
 			msg.DocumentMessage = &waProto.DocumentMessage{
-				Title:         proto.String(mediaPath[strings.LastIndex(mediaPath, "/")+1:]),
-				FileName:      proto.String(mediaPath[strings.LastIndex(mediaPath, "/")+1:]),
+				// outboundFileName, not a manual split on "/": the document
+				// filename travels to the recipient, and on Windows the path
+				// is already backslash-normalised, so the naive split leaks
+				// the whole absolute path. See media_path.go.
+				Title:         proto.String(outboundFileName(mediaPath)),
+				FileName:      proto.String(outboundFileName(mediaPath)),
 				Caption:       proto.String(message),
 				Mimetype:      proto.String(mimeType),
 				URL:           &resp.URL,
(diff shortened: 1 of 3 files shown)


jack-arturo:
Thanks @iB2 — your filename privacy fix is included in [v0.7.0](https://github.com/verygoodplugins/whatsapp-mcp/releases/tag/v0.7.0). Appreciate the contribution and your patience while we caught up on the backlog!

PR #220, opened 2026-09-03
fix(bridge): preserve message ID in text webhooks

TurboKach:
## Problem

Text-only inbound messages were the one webhook payload shape that dropped WhatsApp's native message ID. `WebhookPayload` already carries `MessageID` (`json:"messageId,omitempty"`), and the media and reaction senders already populate it — but `SendWebhook` left it empty, so text webhooks went out with no `messageId` field at all.

That matters because WhatsApp can deliver the same native event more than once, and each delivery arrives over a **different HTTP request**. A consumer that keys idempotency on transport-level headers (`X-Request-ID` and friends) sees two distinct requests and processes the message twice. Without a stable identifier in the body, there is nothing for the consumer to deduplicate on.

## Change

- `webhook.go` — add `SendWebhookWithMessageID(...)`, which serializes `WebhookPayload.MessageID`. `SendWebhook(...)` is kept as a thin wrapper that passes an empty ID, so every existing caller and any external user of the exported function compiles and behaves exactly as before.
- `main.go` — the text-only branch of `handleMessage` now calls `SendWebhookWithMessageID(..., msg.Info.ID)`, bringing text payloads in line with the media and reaction paths.
- `webhook_test.go` — regression test asserting the serialized text webhook JSON carries `messageId`.

Stable native message IDs let webhook consumers safely deduplicate repeated WhatsApp events: the ID is the same across re-deliveries of one message and different between distinct messages, which is exactly the property transport-level request IDs lack.

## Test

```
$ go test ./...
ok  	whatsapp-client	0.543s

$ go build -o /tmp/whatsapp-client-verify .
(ok)
```

## Scope

Three files, +33/−2, additive and backward compatible. No change to any existing exported signature.

### DIFF
--- a/whatsapp-bridge/main.go
+++ b/whatsapp-bridge/main.go
@@ -2009,7 +2009,7 @@ func handleMessage(client *whatsmeow.Client, messageStore *MessageStore, msg *ev
 				msg.Info.ID, mediaType, imageMimeType, filename, imageDownloadPath,
 			)
 		} else {
-			SendWebhook(sender, content, chatJID, msg.Info.IsFromMe, quotedMessageId, quotedSender, quotedContent, quotedIsFromMe, mentionedJIDs)
+			SendWebhookWithMessageID(sender, content, chatJID, msg.Info.IsFromMe, quotedMessageId, quotedSender, quotedContent, quotedIsFromMe, mentionedJIDs, msg.Info.ID)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


jack-arturo:
Thanks @TurboKach — your webhook message-ID fix is included in [v0.7.0](https://github.com/verygoodplugins/whatsapp-mcp/releases/tag/v0.7.0). Appreciate you catching this, and thanks for your patience while we got the release ready!

PR #221, opened 2026-09-04
fix(bridge): persist upload metadata for outbound media

cativo23:
## Summary

- `sendWhatsAppMessage` discarded `url`/`media_key`/`file_sha256`/`file_enc_sha256`/`file_length` when persisting an outbound media message, always passing empty/nil literals to `StoreMessage` instead of the values `client.Upload` had just returned. `downloadMedia`'s completeness check requires all of those fields, so no attachment sent through the bridge (image, video, document) could ever be re-downloaded afterward.
- Confirmed against `messages.db`: every bridge-sent media row had these columns empty, while phone-sent rows (re-synced via the multi-device echo) had them populated. Verified live end-to-end after the fix: send + redownload round-trips a file byte-for-byte.
- Extracted `outboundMediaRow`, a pure function mapping `(mediaPath, whatsmeow.UploadResponse)` to the columns `StoreMessage` persists — the entire surface the bug lived on, now unit-testable without a live client.

## Type of change

- [x] `fix` — bug fix
- [ ] `feat` — new feature
- [ ] `chore` / `docs` / `ci` / `refactor` / `test` / `perf`
- [ ] Breaking change

## Scope check

- [x] This change fits the [`ROADMAP.md`](https://github.com/verygoodplugins/whatsapp-mcp/blob/main/ROADMAP.md) "in scope" list (Bridge stability → media handling), **or** I've opened an issue first to discuss

## Linked issues

None — bug fix with a clear reproduction, per CONTRIBUTING.md this doesn't require a prior issue.

## Testing

- [x] Added or updated tests
- [x] Ran `go test ./... -count=1` — full suite green
- [x] Ran `go vet ./...` and `go build ./...` — clean
- [x] Manually exercised the affected code path (live send + redownload against a real WhatsApp account, byte-for-byte match)

Note: `golangci-lint` isn't installed in my environment, so I couldn't run the CI lint gate locally — `gofmt -l .` is clean and `go vet` passes, but CI's lint run is the first real check of that gate.

Two known gaps, called out rather than hidden:
- A future call site that discards `outboundMediaRow`'s return values in favor of hardcoded literals again isn't caught by the tests added here — that would need a fake `whatsmeow.Client`, which this repo doesn't wire up today. Judged low-probability once the mapping has a name and a single call site.
- The new `DirectPath` consistency check only catches a structural URL/DirectPath disagreement at send time, not a CDN auth token expiring before a later redownload attempt — the same limit `downloadMedia` already has on the inbound path.

## Docs

- [ ] Updated `README.md` (if user-visible) — not needed, no API/behavior contract changed, only internal correctness
- [ ] Updated `AGENTS.md` / `CLAUDE.md` — not needed
- [ ] Updated tool descriptions — not needed, no MCP-facing change
- [ ] Updated `.env.example` — not needed, no new env vars

## Risk / rollback

Low risk: the change only affects what gets written to the `messages` table's media columns for outbound sends and adds a `default` case to a switch that's unreachable today. No schema change, no new dependencies. Revertable with a single `git revert`.

### DIFF
--- a/whatsapp-bridge/main.go
+++ b/whatsapp-bridge/main.go
@@ -1129,6 +1129,21 @@ func classifyMediaPath(mediaPath string) (whatsmeow.MediaType, string, string) {
 	}
 }
 
+// outboundMediaRow derives the columns sendWhatsAppMessage persists for an
+// outbound message: the media category (via classifyMediaPath, so it can't
+// drift from the upload-side classification), the filename, and the upload
+// metadata downloadMedia needs to redownload it later. Returns zero values
+// for a plain text message (mediaPath == ""). Split out as its own function
+// so it's unit-testable without a live whatsmeow.Client.
+func outboundMediaRow(mediaPath string, upload whatsmeow.UploadResponse) (mediaType, filename, url string, mediaKey, fileSHA256, fileEncSHA256 []byte, fileLength uint64) {
+	if mediaPath == "" {
+		return "", "", "", nil, nil, nil, 0
+	}
+	_, _, mediaType = classifyMediaPath(mediaPath)
+	filename = filepath.Base(mediaPath)
+	return mediaType, filename, upload.URL, upload.MediaKey, upload.FileSHA256, upload.FileEncSHA256, upload.FileLength
+}
+
 func buildDisappearingMode() *waProto.DisappearingMode {
 	return &waProto.DisappearingMode{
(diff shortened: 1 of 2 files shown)


jack-arturo:
Thanks @cativo23 — [v0.7.0](https://github.com/verygoodplugins/whatsapp-mcp/releases/tag/v0.7.0) includes your outbound attachment metadata fix, making newly sent attachments retrievable through the MCP. Appreciate the contribution and your patience!

PR #237, opened 2026-09-11
feat(bridge): skip status media and add WHATSAPP_AUTO_DOWNLOAD_MEDIA opt-out

### LINKED ISSUE #238: feat: don't auto-download status media; allow disabling media auto-download
### Problem

The bridge downloads media for every incoming message as it arrives, including WhatsApp status updates (`status@broadcast`). Statuses expire after 24h and are rarely useful to an MCP client, but their media builds up fast: on one account, `whatsapp-bridge/store/status@broadcast` reached **13 GB** out of a 16 GB `store/`. Neither status media nor auto-download in general can be turned off.

### Use case

Someone running the bridge long-term on a laptop or small VM who only occasionally asks their agent about media. They want text and message metadata synced continuously, with media fetched only when needed, and without disk usage growing with every status their contacts post.

### Which scope bucket does this fit?

Bridge stability

### Proposed approach (optional)

1. Never auto-download or webhook-forward `status@broadcast` messages (still store them).
2. Add `WHATSAPP_AUTO_DOWNLOAD_MEDIA` (default `true`); `false` leaves media to on-demand `/api/download` / `download_media`.

Implemented in #237.

### Alternatives considered

- Periodically deleting `store/status@broadcast` by hand, or with a cron job. This works but fixes the symptom after the fact.
- Filtering statuses out before they're stored. This drops data that some users may want to query.

### Pre-flight

- [x] I read `ROADMAP.md` and believe this fits the project's goals.
- [x] I searched existing issues and didn't find a duplicate.
- [x] I'm willing to help test or implement this (optional but appreciated).

iarc13:
## Problem

The bridge auto-downloads media for every incoming message, including status updates (`status@broadcast`). On a real account this put **13 GB** of short-lived status media in `store/status@broadcast` (out of 16 GB total), and there's no way to turn auto-download off.

## Changes

- **Status updates** (`msg.Info.Chat == types.StatusBroadcastJID`) are still stored in `messages.db`, but they are no longer auto-downloaded or forwarded to the webhook.
- **New `WHATSAPP_AUTO_DOWNLOAD_MEDIA`** (default `true`, so existing behavior is unchanged). Setting it to `false` stops background media downloads, including the fallback after a failed webhook-image download. Media is then fetched on demand through `/api/download` / the `download_media` MCP tool. Images forwarded to the webhook still download synchronously, since the payload needs them.
- Documented in `AGENTS.md`, `README.md` and `.env.example`.

Trade-off when set to `false`: WhatsApp media URLs expire after a while, so media that was never downloaded may later fail to download on demand.

## Tests

- `TestHandleMessage_StatusMediaNotDownloadedOrForwarded`: status image is stored, not downloaded, and not sent to the webhook.
- `TestHandleMessage_AutoDownloadDisabledSkipsBackgroundDownload`: the opt-out skips the background download.

Both fail without the change. `go test ./...` and `go vet ./...` pass locally, and `gofmt` is clean. I didn't run `golangci-lint` locally.

Closes #238

https://claude.ai/code/session_01JyJYKefZEsumGLFuVYC8eK


jack-arturo:
Thanks @iarc13 — the status-media suppression and automatic-download setting are in [v0.7.0](https://github.com/verygoodplugins/whatsapp-mcp/releases/tag/v0.7.0). Appreciate the contribution and your patience while we got this release ready!

Issue #223, opened 2026-09-05
feat: at-rest encryption (or documented mitigation) for whatsapp.db / messages.db


puzne2000:
## Problem

`store/whatsapp.db` and `store/messages.db` are plain, unencrypted SQLite files. `whatsapp.db` holds the whatsmeow session/encryption keys (i.e. the credentials that let the bridge act as a linked device), and `messages.db` holds full message history and contacts. Neither is encrypted at rest, and the `store/` directory is created with `0755` / files with default SQLite perms and `0644` for downloaded media (`main.go:101`, `main.go:2163`, `main.go:2229`).

Anyone with filesystem access to the host — another local account on a shared machine, an unencrypted backup, a synced cloud folder, malware, or a stolen/unencrypted disk — can read all message history in plaintext and, more seriously, exfiltrate the session keys from `whatsapp.db` and impersonate the linked device elsewhere.

By contrast, the bridge's REST API is already hardened well beyond typical local-dev-tool bar (bearer token in a `0600` file, constant-time comparison, Host-header allow-list against DNS rebinding — `auth.go`). The on-disk data doesn't currently get the same level of care.

## Use case

A user runs the bridge on a laptop with FileVault/BitLocker disabled (default on many machines), or on a shared/multi-user workstation, or backs up their home directory to an unencrypted external drive or cloud sync folder. Today, that means their entire WhatsApp history and linked-device session keys are exposed in plaintext to anyone who gets that file.

## Which scope bucket does this fit?

Security (not explicitly itemized under ROADMAP.md's "Security" bullets today, but squarely in that bucket — "auditing of file/path inputs" and safe-by-default binding are already listed as examples of the same class of concern).

## Proposed approach (optional)

Rough options, increasing in effort/invasiveness — opening this issue rather than a PR since it's an architectural/dependency decision:

1. **Cheapest, immediate**: tighten `store/` and media file permissions to `0700`/`0600` (currently `0755`/`0644`), and document in the README that users should place `store/` on an encrypted volume (FileVault/BitLocker/LUKS) as the primary mitigation. No new dependencies, no schema change.
2. **Moderate**: encrypt `store/` at rest via SQLCipher (swap the `sqlite3` driver for a build-tagged SQLCipher-enabled one), with a passphrase sourced from an env var or OS keychain. Changes the Go SQL driver dependency and touches every `sql.Open` call in `main.go`.
[...]

PR #241, opened 2026-09-13
fix(bridge): tighten store and media permissions to owner-only

puzne2000:
## Summary

The bridge's runtime store is created group- and world-readable. On a multi-user machine, any other local account can read the full message history and — more importantly — `whatsapp.db`, which holds the linked-device session credentials.

| Location                       | Before | After |
| ------------------------------ | ------ | ----- |
| `store/` (`NewMessageStore`)   | `0755` | `0700` |
| `store/` (`main`)              | `0755` | `0700` |
| `store/` (`auth.go`, token dir) | `0o755` | `0o700` |
| `store/<chat_jid>/` media dirs | `0755` | `0700` |
| downloaded media files         | `0644` | `0600` |

`.bridge-token` was already `0600`; this brings the rest of the store in line with it.

## Security note

Flagging per `AGENTS.md`'s "security-sensitive changes get extra scrutiny": this only narrows permissions, never widens them, and touches no auth logic, path handling, network binding, or command execution. The only behavioral change is that files and directories the bridge creates are no longer readable by other local users.

## Known limitation

`os.MkdirAll` does not alter the mode of a directory that already exists, so existing installs keep whatever permissions their `store/` was created with until it's recreated or manually `chmod`'d. New installs — and any newly created per-chat media subdirectory on an existing install — get the tighter mode immediately. I left the migration out deliberately: retroactively chmod'ing a user's directory tree on startup felt like a surprising side effect for a change of this size, but happy to add it if you'd prefer.

## Relationship to #223

Defense-in-depth only, and explicitly not a substitute for #223 — the data remains plaintext on disk, so this does nothing against a lost disk, an unencrypted backup, or a cloud-sync client running as the same user. It just closes the other-local-user gap, which needed no new dependency or design decision.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` — all passing
- [ ] `golangci-lint run` — not available in my environment; the diff is permission literals and comments only


jack-arturo:
Thanks @puzne2000 — your permission tightening and runtime-storage docs (#240) are both in [v0.7.0](https://github.com/verygoodplugins/whatsapp-mcp/releases/tag/v0.7.0). Appreciate the care around local data privacy, and your patience while we caught up on the backlog!

Issue #205, opened 2026-08-13
QR pairing never completes on WhatsApp Business account: phone shows "Check your connection", bridge only receives unhandled


rogiervanroon:
### Summary

QR pairing consistently fails on a WhatsApp Business account, even on a fresh v0.6.0 checkout with a clean store. The phone scans the QR successfully but then shows "Check your connection and try again". At the exact moment of every scan, the bridge receives a companion_reg_refresh notification that whatsmeow logs as "Unhandled notification" — the pairing handshake (pair-success/pair-error) never starts. Likely a newer Meta companion-registration flow (staged rollout and/or Business-account specific) that whatsmeow doesn't handle yet.

### Steps to reproduce

1. Fresh checkout: git fetch origin && git reset --hard origin/main (v0.6.0, commit 7f518e2)
2. cd whatsapp-bridge && rm -rf store && go run .
3. Wait for the QR code, then scan it immediately with WhatsApp Business (iPhone, up to date) via Settings → Linked Devices → Link a Device
4. Phone shows "Check your connection and try again"; bridge logs only an unhandled companion_reg_refresh (see logs). Reproduces 100% of the time (5+ attempts, multiple QR refreshes).

### Expected behavior

After scanning the QR code, the pairing handshake completes and the bridge logs "Successfully paired".

### Actual behavior

The phone shows "Check your connection and try again" (Dutch: "Controleer je verbinding en probeer opnieuw"). The bridge receives companion_reg_refresh at each scan moment, whatsmeow marks it "Unhandled notification", and no pairing handshake ever begins.

Ruled out: stale code (fresh v0.6.0 + rm -rf store), whatsmeow version (fails identically on the repo pin and on go get -u → v0.0.0-20260810134348), linked-device limit (1 of 4 in use), phone network (Wi-Fi and 5G, no VPN), automatic clocks on both devices, QR timeout (scanned within seconds), WhatsApp app up to date.

Compare WhiskeySockets/Baileys#2488 (April 2026 protocol regression) and tulir/whatsmeow#361 (Business accounts). Happy to test patches — this account is a reliable repro.

### Bridge version

v0.6.0 (commit 7f518e2, fresh checkout of main)

### MCP server version

v0.6.0 (same checkout; not reached yet — failure occurs during bridge pairing, before the MCP server is involved)

### Operating system

macOS

### AI client

Claude Desktop (intended; not reached — see above)

### Relevant logs

```shell
Waiting for QR code scan...
[...]

Issue #245, opened 2026-09-18
feat(mcp): media is unreachable for clients without filesystem access (view_media + transcribe_audio)


seliem-attia:
### Summary

A client with no filesystem access cannot reach any media in a chat, even though the MCP server has full access to the messages. Both media tools hand back a **local file path**:

- `download_media` returns `{"success": true, "file_path": "/…/image_….jpg"}`
- voice notes carry no text at all, so an audio message is an empty `content` field plus a path

For Claude Desktop, a claude.ai chat, or any other client that only speaks MCP, a path is unusable — it cannot open the file. The result is that images, videos and voice notes are effectively invisible, and the natural reading ("the MCP has access to my WhatsApp, so it can see what was sent") is wrong in a way that is hard to guess from the outside.

This is not a permissions problem and it cannot be solved by granting a directory: the client has no filesystem at all.

### Real-world use case

Analysing a project group chat: a few hundred text messages, ~60 images and videos, and two voice notes that contain the actual briefing. The text came through fine. Everything else was a dead end — the assistant could only report that the media "lie outside my permissions" and ask for the files to be uploaded somewhere else. Copying private WhatsApp media into a third-party cloud drive just to make a local MCP server's own data readable is the wrong trade.

### Proposal

Two small, independent tools. I have both implemented and running locally against v0.6.0 and will open them as separate PRs — happy to reshape either one, or drop them if you would rather solve this differently.

**1. `view_media(message_id, chat_jid, max_dimension=1024)`** — returns the picture as MCP image content instead of a path, using FastMCP's `Image`. Videos return their first frame as a still, which is enough to tell what was sent. Both are downscaled first so one photo cannot flood the context. Audio is rejected with a pointer to the transcript. ffmpeg is used when present and is already an optional documented dependency; without it, small images are returned unchanged.

**2. `transcribe_audio(message_id, chat_jid, force=False)`** — transcribes a voice note locally with whisper.cpp and writes the text into the message's `content` field, which is empty for every audio row. After that the transcript shows up through plain `list_messages` for any client, with no second tool call and no repeated transcription. Nothing leaves the machine. A cached transcript returns in milliseconds; a fresh one took ~2 s for a 30-second note with `large-v3-turbo` on an M-series Mac.

[...]

PR #244, opened 2026-09-18
fix(deps): bump whatsmeow for the companion_reg_refresh pairing flow

seliem-attia:
## Summary

- The pinned whatsmeow (`2026-07-30`) predates WhatsApp's companion-registration change, so a QR scan is answered with a `companion_reg_refresh` notification that is never handled and pairing stalls on every attempt (#205).
- Support for that stage landed in whatsmeow `9ec8f76` (2026-09-09) but shipped with a missing `return` in the `RotateADVSecret` case of `qrchan.go`, which closed the QR channel with an empty event ([tulir/whatsmeow#1267](https://github.com/tulir/whatsmeow/issues/1267)). That was fixed in `60027b07a5fc` "qrchan: add missing return" (2026-09-15), so the bump has to land at or after that commit to help.
- whatsmeow now declares `go 1.26.0`, which forces the module directive and the CI toolchain pins. **That is the only reason the workflows are touched here.**

## Type of change

- [x] `fix` — bug fix

## Scope check

- [x] This change fits the `ROADMAP.md` "in scope" list, **or** I've opened an issue first to discuss — dependency bump for a pairing regression; analysis posted on #205
- [x] PR is focused on one concern (split if not) — kept separate from the bridge-side fix in #243 precisely because this one touches CI
- [x] PR is ≤ ~300 LOC, **or** justified in the description — lockfile and pins only

## Linked issues

Refs #205. Companion to #243, which fixes the bridge-side QR loop. Both are needed for pairing to work: this PR makes the rotated code arrive, #243 makes it visible. Merge order does not matter — #243 is inert without this bump, and this bump alone still leaves the rotated code unrendered.

## Testing

- [ ] Added or updated tests — none needed; dependency and pin change only
- [ ] Ran `uv run pytest -v` (Python changes) — no Python changes
- [x] Ran `golangci-lint run` and `go build ./...` (Go changes) — see note below
- [x] Manually exercised the affected code path

`go build ./...`, `go vet ./...`, `go test ./...` (all existing tests) and a `GOOS=linux GOARCH=amd64` build are clean against the new pin. Verified end to end on a real account together with #243: pairing completed, history sync ran, REST API came up.

`golangci-lint` v2.7.1 could not run locally — it fails to decode Go 1.27 export data and reports `typecheck` errors against the standard library, unrelated to this change.

## Docs

- [ ] Updated `README.md` (if user-visible) — the user-visible troubleshooting note ships with #243
[...]

PR #243, opened 2026-09-18
fix(bridge): render every rotated pairing QR code

seliem-attia:
## Summary

- WhatsApp's 2026 companion-registration flow answers a QR scan with a `companion_reg_refresh` notification; whatsmeow rotates the ADV secret and emits a **fresh** code, and only that rotated code can still complete the handshake.
- The QR loop latched on the first code (`qrCodeShown`) and discarded every later one, so a stale code stayed on screen and each scan was validated against a secret the server had already dropped — the phone reports "check your connection", and neither `pair-success` nor `pair-error` ever reaches the bridge.
- Extracts the loop into `renderPairingQRCodes` so it is testable, renders every code (later ones announced as `QR code refreshed`), and adds the first tests for the pairing path.

## Type of change

- [x] `fix` — bug fix

## Scope check

- [x] This change fits the `ROADMAP.md` "in scope" list, **or** I've opened an issue first to discuss — reliability fix for the Go bridge; analysis posted on #205
- [x] PR is focused on one concern (split if not)
- [x] PR is ≤ ~300 LOC, **or** justified in the description — 123 insertions

## Linked issues

Refs #205

## Testing

- [x] Added or updated tests
- [ ] Ran `uv run pytest -v` (Python changes) — no Python changes
- [x] Ran `golangci-lint run` and `go build ./...` (Go changes) — see note below
- [x] Manually exercised the affected code path

`TestRenderPairingQRCodes_RendersEveryCode` is the regression guard: it fails against the old latching behaviour (`rendered 1 code(s), want 2`) and passes with the fix. `TestRenderPairingQRCodes_Outcomes` covers timeout, channel-closed and unknown events.

Verified on a real account: before the change, pairing failed on every attempt with the signature below; after it, the bridge printed the refreshed code, pairing completed (`Successfully paired`), history sync ran and the REST API came up.

```
[Client/Recv DEBUG] <notification ... type="companion_reg_refresh"><companion_reg_refresh/></notification>
[Client/QRChannel DEBUG] Rotating ADV secrets in remaining QR codes
[Client/QRChannel DEBUG] Emitting QR code https://wa.me/settings/linked_devices#...
(no new QR block printed — the latch swallowed it)
```

[...]

PR #246, opened 2026-09-18
feat(mcp): add view_media so media is reachable without a filesystem

seliem-attia:
## Summary

- `download_media` returns a local **file path**. A client that speaks MCP but has no filesystem — Claude Desktop, a claude.ai chat — cannot open it, so images and videos are unreachable for it even though the server has full access to the messages. Granting a directory cannot fix it: there is no filesystem to grant.
- `view_media` returns the picture itself as MCP image content. A video returns its first frame as a still, so one code path covers both, and everything is downscaled first because a single phone photo would flood the caller's context.
- Rendering lives in `media_preview.py` so it is testable without real media, and it fails loudly rather than returning nothing — a silent failure looks identical to "there is no media".

## Type of change

- [x] `feat` — new feature

## Scope check

- [x] This change fits the `ROADMAP.md` "in scope" list, **or** I've opened an issue first to discuss — #245, opened first because CONTRIBUTING asks for a use case and tool description before a new MCP tool
- [x] PR is focused on one concern (split if not) — transcription is a separate PR

## Linked issues

Refs #245

## Testing

- [x] Added or updated tests — 10 new tests in `tests/test_media_preview.py`
- [x] Ran `uv run pytest -v` (Python changes) — 103 passed
- [ ] Ran `golangci-lint run` and `go build ./...` (Go changes) — no Go changes

`uv run ruff check .` and `ruff format --check .` are clean. The tests cover the parts that are easy to get wrong: audio is rejected with a pointer to the transcript rather than rendered, a missing file is reported, the scale filter never enlarges, `-frames:v 1` is what makes image and video share one path, the temp file is cleaned up, and an ffmpeg failure surfaces its own reason instead of a generic message.

Exercised against a real group chat: 48 images and 13 videos that were previously unreadable for a filesystem-less client now come back as viewable stills, at roughly 170 KB per image at the default 1024 px.

## Docs

- [x] Updated `README.md` (if user-visible) — documents `view_media` and notes that `download_media` returns a path
- [x] Updated tool descriptions in `whatsapp-mcp-server/main.py` (if MCP tools changed)
- [ ] Updated `.env.example` (if env vars changed) — no new env vars

## Risk / rollback

Additive: no existing tool changes behaviour, and nothing runs unless the new tool is called. The failure modes are explicit dictionaries rather than exceptions, so a client gets a reason it can act on. Two judgement calls worth a look: JPEG re-encoding for everything ffmpeg touches (simple, lossy, but the alternative is returning multi-megabyte originals), and the 4 MB pass-through cap when ffmpeg is absent. Reverting the commit removes the tool and touches nothing else.


jack-arturo:
Thanks @seliem-attia — [v0.7.0](https://github.com/verygoodplugins/whatsapp-mcp/releases/tag/v0.7.0) includes your media previews, transcription (#247), pairing QR fix (#243), and whatsmeow update (#244). Appreciate all the work on this release and your patience while we caught up on the backlog!

seliem-attia:
Thanks a lot, Jack — and thanks for the follow-up commits and for merging everything in the right order so quickly. Great to see it all land in v0.7.0. I'll switch my local setup over to the release and close #205 and #245 now.

PR #247, opened 2026-09-18
feat(mcp): transcribe voice notes locally into the message content

seliem-attia:
## Summary

- An audio message carries no text, so its `content` column is empty. A client that cannot read the filesystem has no way to learn what was said, and even with a path in hand it would need a second server — with the text not persisted anywhere.
- `transcribe_audio` runs whisper.cpp locally and writes the transcript into that empty column. The words then show up through plain `list_messages` for any client, are searchable with the existing message search, and are never transcribed twice. The audio does not leave the machine.
- Self-contained: no new Python dependencies. whisper.cpp and FFmpeg are external tools, and the tool explains what is missing instead of failing obscurely.

## Type of change

- [x] `feat` — new feature

## Scope check

- [x] This change fits the `ROADMAP.md` "in scope" list, **or** I've opened an issue first to discuss — #245
- [x] PR is focused on one concern (split if not) — image viewing is #246

## Linked issues

Refs #245. Companion to #246; the two are independent and can land in either order.

## Testing

- [x] Added or updated tests — 17 new tests in `tests/test_transcription.py`
- [x] Ran `uv run pytest -v` (Python changes) — 110 passed
- [ ] Ran `golangci-lint run` and `go build ./...` (Go changes) — no Go changes

`uv run ruff check .` and `ruff format --check .` are clean.

Two of the tests guard failure modes that already bit me while building this, both of them silent:

- **Apostrophes.** Hand-quoted SQL produced `unrecognized token: "\"` and skipped exactly the rows whose transcript contained an apostrophe — in German speech, a lot of them. The statement is parameter-bound now, and `test_apostrophes_survive` fails against the old approach.
- **A write that does not stick.** The bridge writes to the same database continuously, so a write can lose the race, and a silent miss is indistinguishable from a note that was never transcribed. The row is read back after writing; a lock is retried, a guard that legitimately matched nothing is not, because waiting cannot help.

Exercised on a real account: 82 voice notes transcribed and stored, then read back through `list_messages` from a client with no filesystem access. A stored transcript returns in milliseconds; a fresh one took ~2 s for a 30-second note with `large-v3-turbo` on an M-series Mac.

## Docs

- [x] Updated `README.md` (if user-visible) — tool, requirements, prefix behaviour
- [x] Updated tool descriptions in `whatsapp-mcp-server/main.py` (if MCP tools changed)
- [x] Updated `.env.example` (if env vars changed) — `WHISPER_MODEL`, `WHISPER_LANGUAGE`

## Risk / rollback

Two things deserve a maintainer's judgement rather than mine:

1. **Writing into `content` at all.** It is what makes transcripts reachable for filesystem-less clients and searchable, and audio rows have no text of their own, so only a blank field is ever filled. But it does mean the column no longer holds solely what people sent, which is a product decision. The prefix (`[transcript (whisper …)]`) keeps the two distinguishable, and a human-written caption on an audio message is never overwritten — there is a test for that. If you would rather not touch `content`, a separate `transcripts` table would work, at the cost of `list_messages` and the message search not seeing them.
2. **`WHISPER_MODEL` has no default**, because whisper.cpp models have no conventional location. That makes the tool inert until configured, which I think is better than guessing a path.

`is_audio` and the audio suffix set also exist in #246's `media_preview`; if both land, unifying them is a small follow-up I am happy to do.

Reverting the commit removes the tool and the module; already-written transcripts stay in the database as plain content, harmless and still marked.


jack-arturo:
Thanks for this, Seliem — making voice notes readable and searchable through the existing tools is a useful addition. I'll add a maintainer commit here to scope transcript reads/writes by both message ID and chat JID, and add an optional OpenAI-compatible transcription endpoint while keeping whisper.cpp as the default. We already share a local Parakeet service between several apps, so that small option lets us reuse it. I'll include tests and docs and keep your original commit intact.

jack-arturo:
Maintainer update pushed in a2d2a3b; the original contribution remains intact. Whisper.cpp stays the default, with an optional OpenAI-compatible endpoint, chat-scoped transcript caching, replay persistence, docs, and regression tests. All 125 Python tests and local Go checks pass, and all required GitHub checks are green.

**Babysit stopped — `ci`.** I’m leaving this unmerged and not marking it merge-ready: the unchanged Go vulnerability job cannot install its latest scanner under Go 1.25; the advisory Python dependency scan also reports findings in unchanged dependencies. Details are in the PR description. Independent review has no verdict yet, and live Parakeet recognition remains unverified because the shared service hung/timed out on synthetic audio (the service was restored to healthy). These follow-ups are outside the scoped transcription change.

seliem-attia:
Thanks a lot for picking this up and for the follow-up commit — chat-scoped caching and the provider option both make sense to me.

On the failing Go Vulnerability Check: #244 bumps all workflows (including the `govulncheck` job in `security.yml`) from Go 1.25 to 1.26, which is exactly what `govulncheck@latest` now requires. Merging #244 first should turn that check green here as well. #244 also carries the whatsmeow bump that #243 depends on for the new pairing flow (#205) — without it #243 is inert — so those two might be worth looking at together.

jack-arturo:
#244 and #243 are merged. This branch is updated with both at f91a3856496f6a54298720bc84b1029ecef8a1c3, and all required checks are green—including the formerly failing Go Vulnerability Check. Local revalidation passed all 125 Python tests plus Go tests/race tests/vet/build/lint. The base-update request is complete; leaving this PR open and unmerged for the transcription feature's final review. The existing live Parakeet validation caveat and advisory Python dependency findings remain documented in the description.

jack-arturo (approved):

